PR #11, opened 2015-02-02
shows age and gender of related person

individual-it:
requires https://github.com/openmrs/openmrs-module-uicommons/pull/11

### DIFF
--- a/omod/src/main/webapp/resources/scripts/relationships/relationships.js
+++ b/omod/src/main/webapp/resources/scripts/relationships/relationships.js
@@ -8,11 +8,24 @@ angular.module('relationships', ['relationshipTypeService', 'relationshipService
         $scope.relationshipTypes = [];
 
         $scope.relationships = [];
+        
+        $scope.personDetails = new Array();
 
         $scope.init = function(personUuid, excludeRelationshipTypes) {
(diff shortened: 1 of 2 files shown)


omod/src/main/webapp/resources/scripts/relationships/relationships.js
```diff
@@ -8,11 +8,24 @@ angular.module('relationships', ['relationshipTypeService', 'relationshipService
...
         $scope.relationships = [];
+        
+        $scope.personDetails = new Array();
 
         $scope.init = function(personUuid, excludeRelationshipTypes) {
             $scope.thisPersonUuid = personUuid;
             RelationshipService.getRelationships({ v: 'default', person: personUuid }).then(function(result) {
                 $scope.relationships = result;
+                
+                //get datails about the related person
```
djazayeri: It should not be necessary to do this. You ought to be able to just change the call a few lines above to request

    v: 'full'

This will return a bigger representation of each relationship, including the more details of the related person

djazayeri:
Please create a ticket at issues.openmrs.org and describe the work you're trying to do here. This PR should reference that ticket.
We'll need to see a screenshot or a mockup of what you're trying to achieve here also.

PR #18, opened 2016-03-05
RA-1065 - Added age to the context model

ssmusoke:
Added age so that it can be used in the formShowIf


mogoodrich:
Looks good to me.

djazayeri:
As long as this is consistent with what it will look like when we refactor
to use REST representations of the patient, go ahead and merge.

(I think that would be patient.person.age.)

-Darius (by phone)

ssmusoke:
@djazayeri Age is included in the REST representation https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PersonResource1_8.java#L61

PR #21, opened 2016-07-12
RA-1166: Added deathdate to the person context model

ssmusoke:
So that it can be used as a test to display forms, also matches the REST webservice representation


mogoodrich:
Looks fine to me...  @djazayeri  any objections with this?

wluyima:
Looks fine to me too and is already merged

PR #30, opened 2016-11-04
RA-1244

mks-d:


wluyima:
Did you configure git to handle line endings properly? It's hard to tell the changes in the PR

mks-d:
@wluyima, what do you mean, are you using some specific Git settings?

dkayiwa:
@mks-d did you look into the above travis failure?
Do you also plan to squash the commits into one?
You also seem to have some file formatting changes in your commits. Do you doing formatting from a separate pull request?

mks-d:
@dkayiwa, I don't know much about TravisCI but I think that - simply - there is no .travis.yml on the new 1.7.x branch and so that's the failure. I don't know whose job it is to make sure that repos are ready to plug into the CI infrastructure but I didn't believe it was mine.

Most of the files I changed were not properly indented in the first place, so I just let my editors correct it, but I didn't realise it would make the PR difficult to read.

So here is what I'll do: I will redo the PR without formatting changes and with adding the same .travis.yml as the one in master. Then will follow a re-formatting PR. Would that be ok?

dkayiwa:
@mks-d yes that will be OK.

mks-d:
@dkayiwa, actually, not the same .travis.yml as the one in master, but one pointing to oraclejdk7 (instead of oraclejdk8).
Anyway, this will take me some time, so bear with me, I guess you can cancel the current PR then :-/

PR #40, opened 2016-12-02
Sanitize addr and phone number fields to prevent xss

isears:


omod/src/main/webapp/fragments/patientdashboard/contactInfoInline.gsp
```diff
@@ -4,12 +4,12 @@
...
 <div class="contact-info-inline">
     <span>
-        ${ ui.format(config.patient.personAddress).replace("\n", ", ")}
+        ${ ui.encodeHtmlContent(ui.format(config.patient.personAddress).replace("\n", ", "))}
         <em>${ ui.message("coreapps.person.address")}</em>
     </span>
     <span class="left-margin">
         <span id="coreapps-telephoneNumber">
-            ${config.patient.telephoneNumber ?: ''}
+            ${ ui.encodeHtmlContent(config.patient.telephoneNumber ?: '') }
```
isears: I was able to insert js into the phone number field by crafting a malicious POST request to an existing patient's editSection page. I used Tamper Data (a free Firefox addon) to edit the POST request. The attached image was taken as I was editing the POST request just after hitting 'confirm' on the 'edit contact info' page.

tele-xss

dkayiwa: Since you did this after hitting confirm, would this code then be relevant? :)

isears: No, I believe it is still necessary. The injected js is manifested on the next page (/openmrs/coreapps/clinicianfacing/patient.page). This issue is the same as https://issues.openmrs.org/browse/RA-649 (sub-issue of RA-452). The only difference is that there seems to be some strong client-side validation of the phone number field built in since the issue was reported. As such, the attack presented here: https://issues.openmrs.org/secure/attachment/43023/XSS-confirm.png is no longer possible. The attack presented here: https://issues.openmrs.org/secure/attachment/43024/execution-at-page-load.png is still possible using the methods I have outlined above.

To verify, I have just attempted js injection on the phone number field on two different local versions of the application (one with the code and one without). The bottom line is that I was able to successfully exploit the xss vulnerability in the phone number field when the code was not present, but was unable to exploit when the code was present.

dkayiwa: Oh i see! Thanks for the extensive explanation! :)

isears: Awesome, thanks for merging my first ever pull request!

I think RA-649 can be resolved after this merge.

dkayiwa: @isears i tried to test it out and noticed that on clicking confirm, the alert window shows up. Could you look into it?

isears: Is this xss that appears in the patient registration summary page, as shown in the attached image?

screen shot 2016-12-04 at 11 27 24 am

If so, I had to fix that issue with changes to uicommons. I can submit a pull request for those changes shortly.

PR #41, opened 2016-12-13
Sanitize various fields in patient search to prevent xss/js injection

isears:


isears:
This patch should solve issue RA-545 (https://issues.openmrs.org/browse/RA-545)

dkayiwa:
Thanks @isears for making openmrs more secure! 👍

PR #42, opened 2016-12-16
Sanitize gender in header and refactor sanitization in search widget

isears:


omod/src/main/webapp/fragments/patientsearch/patientSearchWidget.gsp
```diff
@@ -14,22 +14,17 @@
...
-        age:"${ it.age ?: '' }",
-        birthdate:"${ it.birthdate ? ui.escapeJs(ui.encodeHtmlContent(dateFormatter.format(it.birthdate))) : '' }",
-        // it.birthdateEstimated is of type boolean (doesn't need sanitization)
-        birthdateEstimated: ${ it.birthdateEstimated },
-        identifier:"${ it.patientIdentifier ? ui.escapeJs(ui.encodeHtmlContent(it.patientIdentifier.identifier)) : '' }"
-    });
+                lastViewedPatients.push({
+                    uuid:"${ ui.encodeHtml(it.uuid) }",
+                    name:"${ it.personName ? ui.encodeHtml(ui.format(it.personName)) : '' }",
+                    gender:"${ ui.encodeHtml(it.gender) }",
```
dkayiwa: Did you intentionally remove ui.escapeJs()?

isears:
Was using deprecated `ui.escapeJS()` and `ui.encodeHtmlContent()` methods in the last commit, so the majority of this commit was just replacing those methods with `encodeHtml()`

omod/src/main/webapp/fragments/patientsearch/patientSearchWidget.gsp
isears: Yeah, that's intentional. I figured a single call to encodeHtml is better than two calls to escapeJS and then encodeHtmlContent.

dkayiwa:
@isears when i started to work on this, i got tempted to remove ui.escapeJS(). As a result, introduced some bugs in JavaScript, JSON. So i put them all back because ui.escapeJS() does something that the replacement did not. :)

isears:
I see. What about using ui.encodeJavascript() instead of ui.escapeJS()? I was reading this thread on talk last night: https://talk.openmrs.org/t/use-owasp-encoder-for-xss-attacks/4000 and it inspired me to start using the OWASP filter methods.

mogoodrich:
I remember what @dkayiwa is taking about, but forget the details.  @isears it would be worth looking at the commit history for this file for Daniel's prior change and see if you can find the discussion around it.

isears:
@mogoodrich alright, I'm on it.

isears:
So I found the commit where @dkayiwa added ui.escapeJs() here: https://github.com/openmrs/openmrs-module-coreapps/commit/042e5ad58486851ff79195a876f38187a2fd1241

Later escapeJs() was traded out for encodeHtmlContent() here: https://github.com/openmrs/openmrs-module-coreapps/commit/0be6cb0ff22c68fe29ce0829be80519e9645b42a

And finally both escapeJs() and encodeHtmlContent() were added here: https://github.com/openmrs/openmrs-module-coreapps/commit/0be6cb0ff22c68fe29ce0829be80519e9645b42a

Although I could not find any relevant discussions on any of the commits, I will give my understanding of the issue:

Take this vulnerable code as an example:
```
var lastViewedPatients = [];
    <%  if (showLastViewedPatients && !doInitialSearch) {
            lastViewedPatients.each { it -> %>

                lastViewedPatients.push({
                    uuid:"${ it.uuid }",
                    name:"${ it.personName ? ui.format(it.personName) : '' }",
                    gender:"${ it.gender }",
                    age:"${ it.age ?: '' }",
                    birthdate:"${ it.birthdate ? dateFormatter.format(it.birthdate) : '' }",
                    birthdateEstimated: ${ it.birthdateEstimated },
                    identifier:"${ it.patientIdentifier ? it.patientIdentifier.identifier : '' }"
                });

    <%      }
        }%>
```

Since no sanitization in place, any <script> tags that an attacker inputs into one of these fields will results in classic JS injection into HTML. This issue is prevented by encodeHtmlContent(), which html-encodes all '<' and '>' symbols.

In addition, any double-quote symbol that an attacker inputs will lead to a situation like this:
Malicious payload: `bogus"}; alert('xss');`
Resulting JS: `lastViewedPatient[i] = { uuid:"some-uuid", name: "bogus"}; alert('xss'); ...`
Note that the attacker can exploit XSS even without '<' and '>' symbols. This issue is prevented by ui.escapeJs(), which escapes all double-quote symbols.

The reason I removed ui.escapeJs(ui.encodeHtmlContent()) is because the new OWASP-based sanitization method, ui.encodeHtml(), encodes both less-than/greater-than symbols AND double-quotes.

I have tested this method to prevent attacks of the type outlined here: https://issues.openmrs.org/browse/RA-545 as well as classic XSS using HTML <script> tags. 

As far as I can tell the fix I'm proposing doesn't break anything, but obviously it's up to @dkayiwa and @mogoodrich to determine whether this is the way to go. Either way, just let me know and I can make the required changes (if any).

mogoodrich:
I can't find the conversation either... @dkayiwa if things look okay to you, feel free to go ahead and merge.  Just ping me when you do and I will make sure I do a little regression testing on my side.  Thanks @isears!

dkayiwa:
@isears you can find the full conversation here: https://issues.openmrs.org/browse/UIFR-203

isears:
@dkayiwa thanks for digging that up. I'm still not sure I completely understand the need for escapeJs(encodeHtmlContent()) vs. encodeHtml(), but I don't want to rock the boat by trying to fix something that's not broken. 

This PR still contains a needed fix for the XSS vulnerability in the gender field, however, so I will resubmit with just those changes shortly.

PR #44, opened 2016-12-20
RA-1262 - Patient Search Widget should automatically include person a…

wluyima:
https://issues.openmrs.org/browse/RA-1262 - Patient Search Widget should automatically include person attributes specified in the patient.listingAttributeTypes global property

### DIFF
--- a/omod/src/main/webapp/fragments/patientsearch/patientSearchWidget.gsp
+++ b/omod/src/main/webapp/fragments/patientsearch/patientSearchWidget.gsp
@@ -9,7 +9,10 @@
     ui.includeJavascript("uicommons", "moment-with-locales.min.js")
 %>
 <script type="text/javascript">
-
+    var attributeTypes = [];
+    <% listingAttributeTypeNames.each { %>
+        attributeTypes.push('${ it }');
+    <% } %>
     var lastViewedPatients = [];
     <%  if (showLastViewedPatients && !doInitialSearch) {
             lastViewedPatients.each { it -> %>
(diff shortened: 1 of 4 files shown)


ssmusoke:
@wluyima  For a baby born Dec 1, 2016 the search shows 11 months see attachment 
screenshot 2016-12-20 19 30 26

omod/src/main/webapp/fragments/patientsearch/patientSearchWidget.gsp
```diff
@@ -29,8 +37,11 @@
         // it.birthdateEstimated is of type boolean (doesn't need sanitization)
         birthdateEstimated: ${ it.birthdateEstimated },
         identifier:"${ it.patientIdentifier ? ui.escapeJs(ui.encodeHtmlContent(it.patientIdentifier.identifier)) : '' }"
-    });
-
+    }
+        <% listingAttributeTypeNames.each { attributeName -> %>
+            patientObj["${ attributeName }"] = "${ it.getAttribute(attributeName)?:'' }";
```
isears: It should be noted that showing attributes without any JS-escaping or HTML-encoding causes an XSS vulnerability in the patient search widget (I was able to verify this locally). Unfortunately, the fix will not be as simple as calling ui.escapeJs(ui.encodeHtmlContent()) as is done for name and gender fields since it.getAttribute(attributeName) is not of type String.

isears: @dkayiwa I think something like `ui.encodeHtml(String.valueOf(it.getAttribute(attributeName)))` could work, but it's up to @wluyima to make the final call. I have not extensively tested this solution.

PR #49, opened 2017-01-17
RA-977: Option for user to select Visit Type on Start Visit on New UI

andrewmoko:


omod/src/main/webapp/pages/patientdashboard/patientDashboard.gsp
```diff
@@ -42,7 +42,7 @@
 
 <% if (includeFragments) {
     includeFragments.each { %>
-        ${ ui.includeFragment(it.extensionParams.provider, it.extensionParams.fragment, it.extensionParams.fragmentConfig) }
+        ${ ui.includeFragment(it.extensionParams.provider, it.extensionParams.fragment, [ patient: patient]) }
```
mogoodrich: We can't remove the it.extensionParams.fragmentConfig here as custom fragments may include there own configurations.  You'll have to merge in the patient param and pass all along.  I believe something like this will work:

```
includeFragments.each { 
    var params = it.extensionParams.fragmentConfig ?: {}    // create a base map from the fragmentConfig if it exists, otherwise just create an empty map
    params.patient = patient   // add the patient to the map
%>
       ${ ui.includeFragment(it.extensionParams.provider, it.extensionParams.fragment, params) }
```

I *think* that should do it... I quickly tested something a little less robust, can you test this one out yourself?

omod/src/main/webapp/pages/clinicianfacing/patient.gsp
```diff
@@ -24,7 +24,7 @@
 
 <% if(includeFragments){
     includeFragments.each{ %>
-        ${ ui.includeFragment(it.extensionParams.provider, it.extensionParams.fragment, it.extensionParams.fragmentConfig)}
+        ${ ui.includeFragment(it.extensionParams.provider, it.extensionParams.fragment, [ patient: patient])}
```
mogoodrich: Same issue here as in my other comment below.

omod/src/main/webapp/pages/patientdashboard/patientDashboard.gsp
andrewmoko: Thanks I will

PR #50, opened 2017-01-19
RA-1134 - Add a Register Patient button to the Find Patients page

ssmusoke:
add a register patient button to the search widget


omod/src/main/webapp/fragments/patientsearch/patientSearchWidget.gsp
```diff
@@ -91,3 +91,10 @@
 </form>
 
 <div id="patient-search-results"></div>
+
```
wluyima: Is this on top of the results table? I would think this button should be at the top of the results table

ssmusoke: @wluyima the button is at the bottom of the results table - see discussion here https://issues.openmrs.org/browse/RA-1134

djazayeri: It should be below the table. The idea is that the user looks at the table of results (from top to bottom), doesn't see the result they're looking for, and their eyes end up at the bottom with this "didn't find it?" button.

omod/src/main/webapp/fragments/patientsearch/patientSearchWidget.gsp
```diff
@@ -91,3 +91,10 @@
...
+<% if (showRegisterPatientLink == "true") { %>
+<div>
+${ ui.message("coreapps.findPatient.registerPatient.label") }&nbsp;&nbsp;<a id="patient-search-register-patient" class="button" href="${ui.pageLink("registrationapp", "registerPatient", ["appId":"referenceapplication.registrationapp.registerPatient"])}">${ui.message("registrationapp.registration.label")}</a>
```
djazayeri: This is problematic; you have a reference to a referenceapplication app, but the coreapps module may also be used without the reference application. (E.g. PIH does this.)

Didn't we have a discussion about this, and the good-enough approach is to introduce a global property for "what is the main register patient app"?

ssmusoke: @djazayeri There is no "main" register patient app - unless this is the one registrationapp.basicRegisterPatient 

Also see https://talk.openmrs.org/t/managing-app-definitions-and-extensions-in-openmrs/9601

djazayeri: I was referring to this discussion: https://talk.openmrs.org/t/how-to-make-the-reference-application-more-configurable/7632/29

It's not okay for the default behavior given by the coreapps module to refer to the referenceapplication module (and technically speaking, you can use the coreapps module without the registrationapp module, so even that reference is problematic).

Thinking about this a bit more, here's what I would do. Instead of controlling this by a global property, instead it should be controlled by an app configuration property called "showRegistrationLink" whose value is an application link.

Then, we'd do the same thing that's currently done with Register Patient, e.g. coreapps defines a basic version of the app, but the referenceapplication disables this and provides its own version that has the link to registration turned on.

Does that make sense?

ssmusoke:
@djazayeri I agree with the approach which I think makes alot of sense. 

I have a couple of questions:

1. I have added a "showRegistrationLink" property to the app template, but I do not see how to add it to the app definition JSON

2. What is a better default value null or empty string

3. When creating a new custom app what is the better, recommended or workable approach:

a) Extend the app template creating a new app with all the required details 

b) Extending the coreapps.findPatient app and just adding the new property (once I find out where and how to add it)

djazayeri:
I have no opinion about whether null or empty string is better. Just as long as it works. (Best if you make the code handle both of those interchangeably.)

As to how to make it work, it should be exactly like is done with registrationapp (just substituting coreapps module for registrationapp module in these examples):

* template define in registrationapp, defines config properties: https://github.com/openmrs/openmrs-module-registrationapp/blob/master/omod/src/main/resources/apps/registrationappAppTemplates.json
* simple, dependency-free app instance in registrationapp: https://github.com/openmrs/openmrs-module-registrationapp/blob/master/omod/src/main/resources/apps/registrationapp_app.json
* complex instance in referenceapplication, that defines things in the `config` field: https://github.com/openmrs/openmrs-module-referenceapplication/blob/master/omod/src/main/resources/apps/refapp_registrationapp_app.json

ssmusoke:
@djazayeri Thank you for the pointers, PR updated ...

djazayeri:
At a glance this looks fine to me. If you're tried this both with and without setting that field, I'm satisfied.

omod/src/main/java/org/openmrs/module/coreapps/page/controller/findpatient/FindPatientPageController.java
```diff
@@ -28,6 +31,7 @@ public void get(PageModel model, @RequestParam("app") AppDescriptor app, UiSessi
         model.addAttribute("heading", app.getConfig().get("heading").getTextValue());
         model.addAttribute("label", app.getConfig().get("label").getTextValue());
         model.addAttribute("showLastViewedPatients", app.getConfig().get("showLastViewedPatients").getBooleanValue());
+		model.addAttribute("showRegistrationLink", app.getConfig().get("showRegistrationLink").getTextValue());
```
wluyima: I would think this is a boolean value, not so?

wluyima: Otherwise it won't work unless the value is probably in double quotes

ssmusoke: @wluyima It is a text value - the appId of the registration app instance to link to

wluyima: Well, _showRegistrationLink_ gives the impression that it takes a boolean value, it would be nice to rename to some thing more intuitive that implies its an appId

ssmusoke: @wluyima What do you think of registrationButtonAppId

wluyima: What is the value of this attribute? Is it a URL or just an appId?

ssmusoke: It is an appId

wluyima: I thought the appId is specified via a GP, and this attribute is meant to be used to specify whether to show the create new patient button or not

wluyima: Are you saying that now there is 2 places to define whether to show the register button or not?

ssmusoke: @wluyima There is only one place, in the registrationButtonAppId configuration of the Find Patient app. If this empty or null, then the button is not displayed.

I am thinking that there is no need to update registration core - since this app configuration option solves all problems

wluyima: But from what you are saying, it appears like two, you're saying that the value of the showRegistrationLink attribute is an appId, then there is the GP, see what am saying? Unless showRegistrationLink is a boolean to determine whether to display the registration button or not, do you get my confusion?

ssmusoke: @wluyima With the change to an app configuration parameter there is no need to have a GP. 

If the value is empty or null, then the button is not shown since the button will be broken. 

Hence my recommendation is to rename the configuration variable to registrationButtonAppId

wluyima: As long as you got rid of the GP in registrationapp, that's fine, what was confusing me is that the other PR still exists and it contains the GP

wluyima: So the name of the attribute can be something like _registrationAppId_

ssmusoke: @wluyima That is not clear since one would wonder what it does i am thinking mymodule.findPatient.registrationAppId

wluyima: The 'schema' of an app config  is really the [template](https://github.com/openmrs/openmrs-module-coreapps/blob/master/omod/src/main/resources/apps/coreappsAppTemplates.json), there will be no wondering as long as the field is well documented in the template

wluyima: So please remember to add the new field to the template

djazayeri: Just seeing this now...wouldn't it be better if it _is_ a URI? e.g. "registerPatientLink".

wluyima: @djazayeri apparently it's an appId and not a URI as such

ssmusoke: Fixed the name of the parameter, and detailed usage in the app template

djazayeri:
I see I'm too late now, but what I meant to say was, it would be better behavior to merely specify a URL, instead of building in the requirement that the "Add New Patient" link must be an instance of the registrationapp's register patient module.

This implementation won't allow me to use a completely custom register patient screen.
Whereas if you were just passing a URL, you can also give the URL for `...registerPatient.page?appId=...`

wluyima:
Makes sense that one might want it to be a URL vs just an appId

wluyima:
Unfortunately, I don't seem to have privileges to reopen RA tickets

ssmusoke:
@djazayeri I get it now. Will reopen the ticket and submit a new PR tomorrow with a link instead of just an appId

ssmusoke:
@djazayeri Thinking about this a lot more if I am to specify a url say registrationapp/registerPatient.page?appId=registrationapp.basicRegisterPatient then the question is how can I ensure that this url is always well formed with respect to the context path. 

There are two options:

1. Add provider and resource parameters to generate a proper url (harder but will ensure that the code follows existing conventions)

2. Use the context method to prepend to the provided url to ensure that it is not broken (simple and fixes the issue for now)

wluyima:
The injects into the page model for use in any GSP, you can access it as _${contextPath}_, it's also available for use in javascript via a global variable named _OPENMRS_CONTEXT_PATH_

With that said, I'd think it's always better to let a developer provide a provide and resource and use the utility function in the framework that generates links i.e `ui.resourceLink(provider, resource) ` instead of taking a full URL.

djazayeri:
Actually it's better to just pass a URL. The provider + resource is
specific to the UI Framework, and we want to make sure this could also like
to an OWA-provided page.

But you should require the admin to put in a URL relative to contextPath,
and you can add that on.

PR #52, opened 2017-01-26
RA-977: Option for user to select Visit Type on Start Visit on New UI

andrewmoko:


mogoodrich:
Thanks @andrewmoko I will take a look!

PR #63, opened 2017-02-27
RA-1301 Dashboard Template Widget - WIP

AdamGrzybkowski:


omod/src/main/resources/apps/coreappsAppTemplates.json
```diff
@@ -94,6 +94,18 @@
                 "defaultValue": ""
             }
         ]
+    },
+    {
+        "id": "coreapps.template.baseTemplate",
```
AdamGrzybkowski: @rkorytkowski What do you think about this ? We can't set in the template fields like icon or label like we could in app description and this is the cleanest way  I've found without modifying  [this](https://github.com/openmrs/openmrs-module-appframework/blob/master/api/src/main/java/org/openmrs/module/appframework/domain/AppTemplate.java)

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/BaseFragmentController.java
```diff
@@ -0,0 +1,12 @@
+package org.openmrs.module.coreapps.fragment.controller;
+
+import org.openmrs.module.appframework.domain.AppDescriptor;
+import org.openmrs.ui.framework.fragment.FragmentConfiguration;
+
+abstract public class BaseFragmentController {
+
+    public void controller(FragmentConfiguration config, AppDescriptor app) {
+        config.addAttribute("icon", app.getConfig().get("icon").toString());
```
AdamGrzybkowski: Should I check if the app is instance of the coreapps.template.baseTemplate ?

PR #70, opened 2017-03-07
RA-1306 Visits By Encounter Type Widget

AdamGrzybkowski:
I wasn't sure what to do if there were no encounterTypes for a visit. See below for my solution. Second thing is that for now clicking the visit date redirects you only to all patient visits not specific one. To make this we need to provide visit and patient id which cannot be done via REST (?) so we would have to modify fragment controller ?

Screen shot
ra-1306

### DIFF
--- a/omod/src/main/resources/apps/dashboardWidgets_app.json
+++ b/omod/src/main/resources/apps/dashboardWidgets_app.json
@@ -47,5 +47,30 @@
         }
       }
     ]
+  },
+  {
+    "id": "coreapps.visitByEncounterType",
+    "instanceOf": "coreapps.template.dashboardWidget",
+    "description": "coreapps.visitByEncounterType.app.description",
+    "order": 10,
+    "config": {
+      "widget": "visitbyencountertype",
+      "icon": "icon-group",
+      "label": "Recent Visits",
+      "maxRecords": "3",
+      "maxAge": "1d",
+      "combineEncounterTypes": "true"
+    },
+    "extensions": [
+      {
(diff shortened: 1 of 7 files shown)


ssmusoke:
@AdamGrzybkowski I think its okay to leave the encounter types blank if there are none for a visit. 

Can't the visitId and patientId be gotten from the visit via REST? Since I see visit.encounterType I am assuming that you can have visit.patientId and visit.id (please excuse my ignorance of REST here)

rkorytkowski:
@ssmusoke, REST doesn't expose IDs, just UUIDs. @AdamGrzybkowski, we need to make the page you link under dates to accept UUIDs in addition to IDs.

ssmusoke:
@rkorytkowski Got you - my understanding has been that the patientId takes both an Id and uuid in the same parameter throughout so I assumed that it was the same for visit

rkorytkowski:
@AdamGrzybkowski , did you actually verify it doesn't work when you pass uuid instead of id as Stephen suggests? :)

rkorytkowski:
http://qa-refapp.openmrs.org/openmrs//coreapps/patientdashboard/patientDashboard.page?patientId=3282d83b-55c3-44b5-af39-40f599159076&visitId=6795#

AdamGrzybkowski:
@rkorytkowski Yes I verified it but I see you already did it too : )

rkorytkowski:
The fix should be easy. Look around here https://github.com/openmrs/openmrs-module-coreapps/blob/master/omod/src/main/java/org/openmrs/module/coreapps/page/controller/patientdashboard/PatientDashboardPageController.java#L43

PR #75, opened 2017-03-10
RA-1304 - Data Integrity Violations Widget

tmarzeion:
I've used mocked response from future `dataintegrity` REST endpoint to create layout.
Ticket will be ready when endpoint is ready and this app's code is tweaked a bit to fit real-life response.
ra-1304_mockup
Note: Bold `Fix` labels got their `actionUrl` defined, so it means that they are clickable.
@rkorytkowski @ssmusoke Do you think that way of distinguishing clickable and non-clickable `Fix`es is OK?

### DIFF
--- a/omod/src/main/webapp/resources/scripts/fragments/dashboardwidgets/dataintegrityviolations/dataintegrityviolations.component.js
+++ b/omod/src/main/webapp/resources/scripts/fragments/dashboardwidgets/dataintegrityviolations/dataintegrityviolations.component.js
@@ -1,0 +1,8 @@
+angular.module("openmrs-contrib-dashboardwidgets.dataintegrityviolations", [ "openmrs-contrib-uicommons" ]).component('dataintegrityviolations', {
+    template: '<div ng-include="getTemplate()">',
+    controller: DataIntegrityViolationsController,
+    controllerAs: 'ctrl',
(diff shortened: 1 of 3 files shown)


ssmusoke:
@tmarzeion I would suggest that if there is no actionUrl you do not show a fix label. 

I would like to suggest that this batch of widgets be moved to the Data Integrity module as they are a special use case and not always packaged with Ref App. @rkorytkowski your advice needed here

tmarzeion:
@ssmusoke I've replied to Your suggestion about keeping widget inside module at OpenMRS Jira :)
I've also changed layout according to Your opinion, tell me what do You think:
mockup_2

tmarzeion:
I've removed this App from `dashboardWidgets_app.json` since we don't want to load it by default (it is caused because of its dependency of `dataintegrity` module).
Here is example of app config:
http://pastebin.com/77FEFf6r
This PR is dependent by https://github.com/openmrs/openmrs-module-dataintegrity/pull/14

UI:
ra-1304-finish

rkorytkowski:
Please add the app config to the wiki at https://wiki.openmrs.org/display/docs/Patient+Summary+Widget+Documentation

tmarzeion:
@rkorytkowski Done

PR #79, opened 2017-03-28
Fixing java.lang.IllegalArgumentException: The date must not be null

rbuisson:
at org.apache.commons.lang.time.DateUtils.add(DateUtils.java:414) when
we have two active visits as a result of a failure to end the current
visit before starting a new one hence leading to stopDatetime being null


dkayiwa:
Is this only in the 1.7.x branch?

rbuisson:
Yes. This commit is already present on the master branch. I am just back-porting to 1.7.x

PR #83, opened 2017-04-20
RA-1342, implement edit and delete functions

cioan:
Hi @mogoodrich , here is the commit that implements relationships editing via the widget. It also contains a new privilege that gives user access to the edit function. Thanks.


mogoodrich:
I gave this a look over and it looks good to me... we will want to add localization of text, but that can be in a follow-on commit.

@rkorytkowski  or @wluyima  did you want to give this a look over as well?  Otherwise I think we can go ahead and merge this in.  Thanks @cioan!

PR #84, opened 2017-04-21
RA-1342, fix bug when person in relationship is not patient

cioan:
Hi @mogoodrich , thanks for identifying this bug with the person in a relationship that is not a patient.


mogoodrich:
Looks good, thanks @cioan !

PR #86, opened 2017-04-24
RA-1342, add optional relationships type filter and prevent adding mu…

cioan:
…ltiple relationships with the same person

### DIFF
--- a/omod/src/main/webapp/resources/scripts/fragments/dashboardwidgets/relationships/relationships.controller.js
+++ b/omod/src/main/webapp/resources/scripts/fragments/dashboardwidgets/relationships/relationships.controller.js
@@ -263,7 +263,8 @@ function RelationshipsController(openmrsRest, $scope) {
     }
 
     ctrl.searchPersons = function() {
-        if (angular.isDefined(ctrl.searchPerson) && (ctrl.searchPerson.length > ctrl.minSearchLength) ) {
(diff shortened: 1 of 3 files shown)


cioan:
Hi @mogoodrich , here are the fixes for the bugs Dave found. Thanks.

omod/src/main/webapp/resources/scripts/fragments/dashboardwidgets/relationships/relationships.controller.js
```diff
@@ -97,25 +112,41 @@ function RelationshipsController(openmrsRest, $scope) {
...
-                relTypeA.type = "B";
-                ctrl.types.push(relTypeA);
-            }
-            if (findRelTypeByName(type.bIsToA) == null) {
-                var relTypeB = {};
-                relTypeB.uuid = type.uuid;
-                relTypeB.name = type.bIsToA;
-                relTypeB.type = "A";
-                ctrl.types.push(relTypeB);
+            if ((ctrl.allowedTypes.length < 1) || ((ctrl.allowedTypes.length > 0)
```
cioan: Thanks @mogoodrich . I was trying to say that if a relationship type filter was not specified then we display all types, OR if a filter was defined then we only display the types included in the filter. I will add more comments and put the second OR statement on the same line.

PR #95, opened 2017-05-07
RA-981 moving tests to omod submodule

gayanW:


dkayiwa:
@gayanW looking at the commit message, is this about doing what is in RA-981 or is it about moving tests to the omod submodule?

gayanW:
@dkayiwa this is an test PR. see https://talk.openmrs.org/t/adding-webservices-rest-dependencies-to-coreapps-module-causes-tests-to-fail/11279/8?u=gayanw

It include all the changes for RA-981, and not just the moving of tests. However the moving of tests causes tests to fail. I'd appreciate if you could have a look.

dkayiwa:
@gayanW then your commit message should have said so. But the way it currently is, it looks like dealing with only the moving of the tests.

gayanW:
@dkayiwa right. but that's what makes the tests to fail. i thought it would help so you could focus on just that when reviewing.

dkayiwa:
The commit message is all about what is contained in the commit. It has nothing to do with which section of code you want some one to pay special attention to. Do you agree? 😊

gayanW:
Yes I do :D

dkayiwa:
Fixed failing tests with https://github.com/openmrs/openmrs-module-coreapps/commit/3390bd975217687c01651a60bc5d19106dcb31ca

PR #110, opened 2017-06-19
RA-1381, improve the runtime of this page

cioan:
Hi @mogoodrich  and @andrewmoko , I made some changes to this page and its controller to prevent crashing in our production servers. 
@andrewmoko : Do you have a test server that uses the multicolor multi visit type feature? Our PIH EMR and reference app QA servers only have one visit type. Just wanted to make sure I did not break any of the functionality you introduced with the multicolor visit types. Thanks!


mogoodrich:
Thanks @cioan!

PR #123, opened 2017-10-02
RA-1366: Update recent visits widgets

henziger:
This commit removes the old recent visits widget
and updates the current widget to use the correct icon and
removes the restriction of only showing visits that are
more recent than 3 months.

Ref ticket: https://issues.openmrs.org/browse/RA-1366


ssmusoke:
Thanks @henziger

PR #127, opened 2017-10-26
RA-1366: Disable old visits widget in Ref App's clinician dashboard.

henziger:


mogoodrich:
Looks good to me, thanks @henziger!

fyi @ssmusoke....

PR #160, opened 2018-10-29
TRUNK-5343: removed redundant setting of type for an encounter

dsurrao:
https://issues.openmrs.org/browse/TRUNK-5343


dkayiwa:
@dsurrao i actually set the encounter type to fix the NPE that i was getting. So removing it will lead to the exact problem when running on platform 2.2

PR #173, opened 2019-03-14
RA-1575: Set the minSearchCharacters to use the value of the minSearchCharacters global property

ssmusoke:
https://issues.openmrs.org/browse/RA-1575

### DIFF
--- a/omod/src/main/webapp/fragments/patientsearch/patientSearchWidget.gsp
+++ b/omod/src/main/webapp/fragments/patientsearch/patientSearchWidget.gsp
@@ -54,7 +54,7 @@
         var widgetConfig = {
             initialPatients: lastViewedPatients,
             doInitialSearch: ${ doInitialSearch ? "\"" + ui.escapeJs(doInitialSearch) + "\"" : "null" },
-            minSearchCharacters: ${ config.minSearchCharacters ?: 3 },
+            minSearchCharacters: ${ minSearchCharacters ?: 3 },
             afterSelectedUrl: '${ ui.escapeJs(config.afterSelectedUrl) }',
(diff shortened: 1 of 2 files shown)


omod/src/main/webapp/fragments/patientsearch/patientSearchWidget.gsp
```diff
@@ -54,7 +54,7 @@
         var widgetConfig = {
             initialPatients: lastViewedPatients,
             doInitialSearch: ${ doInitialSearch ? "\"" + ui.escapeJs(doInitialSearch) + "\"" : "null" },
-            minSearchCharacters: ${ config.minSearchCharacters ?: 3 },
```
dkayiwa: Did you accidentally change this from config.minSearchCharacters?

mogoodrich: I assumed that this was a bug that @ssmusoke discovered and was fixing in this commit... ie it wasn't picking up the parameter because it should have just been minSearchCharacters... however, admittedly, I didn't double-check this... :)

ssmusoke: @dkayiwa Nope that is the fix which uses the value from the fragment controller, @mogoodrich you are right that is the bug fix - removing the config parameter

ssmusoke: @dkayiwa Thank you for the additional search, only one location, in the patientSearchAndSelectWidget is applicable since its the only GSP. The rest as JS files which use the config values in the gsp

PR #179, opened 2019-03-28
RA-1580, add support for plotting the BMI

cioan:


mogoodrich:
LGTM... this is a great addition!  On one hand, it would be nice to extract out the BMI function to somewhere else, but, on the other hand, it would be such a common use case for graphs, it will be easier to find and for people to use here.

PR #180, opened 2019-04-09
RA-1520 Improve Performance of Latest Obs widget 

ssmusoke:
https://issues.openmrs.org/browse/RA-1520

### DIFF
--- a/omod/src/main/java/org/openmrs/module/coreapps/web/resource/LatestObsResource.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/web/resource/LatestObsResource.java
@@ -13,10 +13,6 @@
  */
 package org.openmrs.module.coreapps.web.resource;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Vector;
-
 import org.openmrs.Concept;
 import org.openmrs.Obs;
 import org.openmrs.Patient;
@@ -31,12 +27,22 @@ import org.openmrs.module.webservices.rest.web.api.RestService;
 import org.openmrs.module.webservices.rest.web.resource.api.PageableResult;
 import org.openmrs.module.webservices.rest.web.resource.impl.EmptySearchResult;
 import org.openmrs.module.webservices.rest.web.resource.impl.NeedsPaging;
+import org.openmrs.module.webservices.rest.web.response.ResponseException;
 import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_11.ObsResource1_11;
 import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8.PatientResource1_8;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Vector;
+
 @Resource(name = RestConstants.VERSION_1 + "/latestobs", order = 1, supportedClass = Obs.class, supportedOpenmrsVersions = {
-		"1.8.*", "1.9.*", "1.10.*", "1.11.*", "1.12.*", "2.0.*", "2.1.*", "2.2.*" })
+		"1.8.*", "1.9.*", "1.10.*", "1.11.*", "1.12.*", "2.0.*", "2.1.*", "2.2.*", "2.3.*" })
 public class LatestObsResource extends ObsResource1_11 {
+
+	@Override
+	protected NeedsPaging<Obs> doGetAll(RequestContext context) throws ResponseException {
+		return new NeedsPaging<Obs>(new ArrayList<Obs>(), context);
+	}
 	
 	@Override
 	protected PageableResult doSearch(RequestContext context) {
--- a/omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
+++ b/omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
@@ -15,39 +15,44 @@ export default class LatestObsForConceptListController {
         this.maxAgeInDays = this.widgetsCommons.maxAgeToDays(this.config.maxAge);
 
         // Remove whitespaces
-        this.config.concepts = this.config.concepts.replace(/\s/g,'');
-
+        let concept_list = this.config.concepts.replace(/\s/g,'');
+        // remove all concepts over the maximum number of concepts
         let concepts = this.config.concepts.split(",");
-        for (let i = 0; i < concepts.length; i++) {
-            let concept = concepts[i];
-            // Fetch last obs for concept
-            this.openmrsRest.list('obs', {
-                patient: this.config.patientUuid,
-                v: 'full',
-                concept: concept
-            }).then((resp) => {
-                // Don't add more items if max concept count is reached
-                if (this.obs.length < this.maxConceptCount && resp.results.length > 0) {
-                    let obs = resp.results[0];
-                    // Don't add obs older than maxAge
-                    if (angular.isUndefined(this.maxAgeInDays) || this.widgetsCommons.dateToDaysAgo(obs.obsDatetime) <= this.maxAgeInDays) {
-                        // Add last obs for concept to list
-
-                        if (['8d4a505e-c2cc-11de-8d13-0010c6dffd0f',
-                                '8d4a591e-c2cc-11de-8d13-0010c6dffd0f',
-                                '8d4a5af4-c2cc-11de-8d13-0010c6dffd0f'].indexOf(obs.concept.datatype.uuid) > -1) {
-                            //If value is date, time or datetime
-                            var date = this.$filter('date')(new Date(obs.value), this.config.dateFormat);
-                            obs.value = date;
-                        } else if (angular.isDefined(obs.value.display)) {
-                            //If value is a concept
-                            obs.value = obs.value.display;
-                        }
-
-                        this.obs.push(obs);
-                    }
-                }
-            });
+        // from the number of elements specified by maxConceptCount remove the remaining elements
+        let elements_to_remove = concepts.length - this.maxConceptCount;
+        if (elements_to_remove > 0) {
+            concepts.splice(this.maxConceptCount, elements_to_remove);
         }
+
+        concept_list = concepts.join(",");
+
+        // Fetch last obs for the list of concepts
+        this.openmrsRest.list('latestobs', {
+            patient: this.config.patientUuid,
+            v: 'full',
+            concept: concept_list
+        }).then((resp) => {
+            // Process the results from the list of concepts as not all of them may have data
+            for (let i = 0; i < resp.results.length; i++) {
+                let obs = resp.results[i];
+                // Don't add obs older than maxAge
+                if (angular.isUndefined(this.maxAgeInDays) || this.widgetsCommons.dateToDaysAgo(obs.obsDatetime) <= this.maxAgeInDays) {
+                    // Add last obs for concept to list
+
+                    if (['8d4a505e-c2cc-11de-8d13-0010c6dffd0f',
+                        '8d4a591e-c2cc-11de-8d13-0010c6dffd0f',
+                        '8d4a5af4-c2cc-11de-8d13-0010c6dffd0f'].indexOf(obs.concept.datatype.uuid) > -1) {
+                        //If value is date, time or datetime
+                        var date = this.$filter('date')(new Date(obs.value), this.config.dateFormat);
+                        obs.value = date;
+                    } else if (angular.isDefined(obs.value.display)) {
+                        //If value is a concept
+                        obs.value = obs.value.display;
+                    }
+
+                    this.obs.push(obs);
+                }
+            }
+        });
     }
 }
--- a/omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.spec.js
+++ b/omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.spec.js
@@ -27,14 +27,24 @@ describe('LatestObsForConceptList', () => {
     });
 	
 	it('should query for concepts from config', () => {
-		let bindings = {config: { maxAge: '2w', concepts: 'uuid-1, uuid-2', patientUuid: 'patientUuid' }};
+		let bindings = {config: { maxAge: '2w', concepts: 'uuid-1,uuid-2', patientUuid: 'patientUuid' }};
 		let ctrl = $componentController('latestobsforconceptlist', {$scope}, bindings);
-		
-		$httpBackend.expectGET('/ws/rest/v1/obs?concept=uuid-1&patient=patientUuid&v=full').respond({results: []});
-		$httpBackend.expectGET('/ws/rest/v1/obs?concept=uuid-2&patient=patientUuid&v=full').respond({results: []});
-		
+
+		$httpBackend.expectGET('/ws/rest/v1/latestobs?concept=uuid-1,uuid-2&patient=patientUuid&v=full').respond({results: []});
+
 		ctrl.$onInit();
 		
 		$httpBackend.flush();
     });
+
+	it('should query for a single concept from config', () => {
+		let bindings = {config: { maxAge: '2w', concepts: 'uuid-1', patientUuid: 'patientUuid' }};
+		let ctrl = $componentController('latestobsforconceptlist', {$scope}, bindings);
+
+		$httpBackend.expectGET('/ws/rest/v1/latestobs?concept=uuid-1&patient=patientUuid&v=full').respond({results: []});
+
+		ctrl.$onInit();
+
+		$httpBackend.flush();
+	});
 });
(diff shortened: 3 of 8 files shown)


pom.xml
```diff
@@ -52,7 +52,7 @@
         <idgenVersion>4.4.0</idgenVersion>
 		<metadatamappingVersion>1.3.2</metadatamappingVersion>
         <uicommonsVersion>2.3.0</uicommonsVersion>
-        <webservicesRestVersion>2.11</webservicesRestVersion>
+        <webservicesRestVersion>2.23</webservicesRestVersion>
```
dkayiwa: Change the above to 2.23.0

dkayiwa:
In the root pom.xml file, change from <type>tests-jar</type> to <classifier>tests</classifier>

Then in the omod's pom.xml file, add <classifier>tests</classifier>to the webservices.rest-omod-common dependency

ssmusoke:
Daniel I am still working on this one

dkayiwa:
Oh sorry i thought it was a throw away pr to only show me the error. 😊 
I have just reopened it.

ssmusoke:
@dkayiwa if you get a chance, I am get an exception loading ApplicationContext during tests when running `mvn clean install`

ssmusoke:
@dkayiwa if you get a chance, I am get an exception loading ApplicationContext during tests when running `mvn clean install`

dkayiwa:
Is everything committed?

dkayiwa:
Is everything committed?

ssmusoke:
@dkayiwa - yes all files have been committed

dkayiwa:
@ssmusoke add empty classes org.openmrs.AllergyReaction and org.openmrs.Allergen in the test folder of the omod project

ssmusoke:
@dkayiwa Looks like the issue is loading the WebTest Helper see stack trace https://gist.github.com/ssmusoke/90f25648d548f1633a098f87f7fcd411

omod/src/test/java/org/openmrs/AllergyReaction.java
```diff
@@ -0,0 +1,4 @@
+package org.openmrs.module.coreapps;
```
dkayiwa: The package should be org.openmrs

ssmusoke: Thanks @dkayiwa

ssmusoke:
@dkayiwa was doing final tests to validate functionality 

https://issues.openmrs.org/browse/RA-1520

ssmusoke:
@mogoodrich This may be of interest to you if you are using this dashboard widget to display data

ssmusoke:
@mogoodrich @cioan ping for your input and feedback

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.spec.js
```diff
@@ -30,9 +30,8 @@ describe('LatestObsForConceptList', () => {
 		let bindings = {config: { maxAge: '2w', concepts: 'uuid-1, uuid-2', patientUuid: 'patientUuid' }};
 		let ctrl = $componentController('latestobsforconceptlist', {$scope}, bindings);
 		
-		$httpBackend.expectGET('/ws/rest/v1/obs?concept=uuid-1&patient=patientUuid&v=full').respond({results: []});
```
dkayiwa: Did you intentionally remove these two lines?

omod/src/test/java/org/openmrs/Allergen.java
```diff
@@ -0,0 +1,4 @@
+package org.openmrs;
+
```
dkayiwa: Can you add some comments for why this class was added? Will be very useful to even yourself when you, after some years, ask yourself, why the hell you added this. 😊

omod/src/test/java/org/openmrs/AllergyReaction.java
```diff
@@ -0,0 +1,4 @@
+package org.openmrs;
+
```
dkayiwa: Same as above.

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.spec.js
ssmusoke: @dkayiwa Yes I did since the message signature has changed to support two concepts, but now that you mention it, I should have just added the third test since the other two should still work.

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
```diff
@@ -15,39 +14,35 @@ export default class LatestObsForConceptListController {
         this.maxAgeInDays = this.widgetsCommons.maxAgeToDays(this.config.maxAge);
 
         // Remove whitespaces
-        this.config.concepts = this.config.concepts.replace(/\s/g,'');
-
-        let concepts = this.config.concepts.split(",");
-        for (let i = 0; i < concepts.length; i++) {
-            let concept = concepts[i];
-            // Fetch last obs for concept
```
dkayiwa: Is this just a formatting change? It is hard to tell what exactly changed.

ssmusoke: @dkayiwa The whole loop changed, removed an outer loop and how it works, so instead of fetching one concept at a time, we process the results for multiple concepts at a go

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.spec.js
ssmusoke: I have added a test case for a single concept instead

ssmusoke:
@dkayiwa Thanks for the feedback all resolved and PR updated

omod/src/test/java/org/openmrs/Allergen.java
```diff
@@ -0,0 +1,7 @@
...
+/**
+ * Added to ensure that the tests pass
```
dkayiwa: I comment like this would be more appropriate: https://github.com/openmrs/openmrs-module-idgen/blob/master/omod/src/test/java/org/openmrs/CodedOrFreeText.java#L12

ssmusoke: @dkayiwa Done

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.spec.js
```diff
@@ -27,14 +27,24 @@ describe('LatestObsForConceptList', () => {
...
-		
+
+		$httpBackend.expectGET('/ws/rest/v1/latestobs?concept=uuid-1,uuid-2&patient=patientUuid&v=full').respond({results: []});
+
 		ctrl.$onInit();
 		
 		$httpBackend.flush();
     });
+
+	it('should query for a single concept from config', () => {
```
dkayiwa: Shouldn't you have some assertions at the end of this method?

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.spec.js
```diff
@@ -27,14 +27,24 @@ describe('LatestObsForConceptList', () => {
     });
 	
 	it('should query for concepts from config', () => {
-		let bindings = {config: { maxAge: '2w', concepts: 'uuid-1, uuid-2', patientUuid: 'patientUuid' }};
+		let bindings = {config: { maxAge: '2w', concepts: 'uuid-1,uuid-2', patientUuid: 'patientUuid' }};
 		let ctrl = $componentController('latestobsforconceptlist', {$scope}, bindings);
```
dkayiwa: Aren't we supposed to be backwards compatible?

omod/src/test/java/org/openmrs/Allergen.java
```diff
@@ -0,0 +1,16 @@
...
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs;
+
+//TODO this dummy class is just for the sake of making LatestObsRestControllerTest pass.
+//it fails because the platform dependency version currently used, does not have this class.
+//after this has been solved, this dummy class can be deleted.
```
brandones: Maybe these should link to a JIRA ticket?

ssmusoke: I had forgotten to add the JIRA ticket in the PR description

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.spec.js
ssmusoke: @dkayiwa This comment is not clear - backwards compatible in what way?

omod/src/test/java/org/openmrs/module/coreapps/web/controller/LatestObsRestControllerTest.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public long getAllCount() {
+        return 0;
+    }
+
+    @Test
+    public void testSearchForSingleConcept() throws Exception {
+        // patientuuid - 5946f880-b197-400b-9caa-a3c661d23041, patientId - 7, patient uuid concepts - 11716f9c-1434-4f8d-b9fc-9aa14c4d6126 from standardDataSet
```
dkayiwa: What is the use of the above comment?

omod/src/main/java/org/openmrs/module/coreapps/web/resource/LatestObsResource.java
```diff
@@ -31,12 +27,22 @@
...
+import org.openmrs.module.webservices.rest.web.response.ResponseException;
 import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_11.ObsResource1_11;
 import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8.PatientResource1_8;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Vector;
+
 @Resource(name = RestConstants.VERSION_1 + "/latestobs", order = 1, supportedClass = Obs.class, supportedOpenmrsVersions = {
 		"1.8.*", "1.9.*", "1.10.*", "1.11.*", "1.12.*", "2.0.*", "2.1.*", "2.2.*" })
```
dkayiwa: Can we also add platform 2.3?

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
```diff
@@ -15,39 +14,35 @@ export default class LatestObsForConceptListController {
...
+        let concept_list = this.config.concepts.replace(/\s/g,'');
+
+        // Fetch last obs for the list of concepts
+        this.openmrsRest.list('latestobs', {
+            patient: this.config.patientUuid,
+            v: 'full',
+            concept: concept_list
+        }).then((resp) => {
+            // Process the results from the list of concepts as not all of them may have data
+            for (let i = 0; i < resp.results.length; i++) {
```
dkayiwa: Did you intentionally ignore the this.maxConceptCount check?

omod/src/test/java/org/openmrs/module/coreapps/web/controller/LatestObsRestControllerTest.java
ssmusoke: Removed

omod/src/main/java/org/openmrs/module/coreapps/web/resource/LatestObsResource.java
ssmusoke: Done

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
ssmusoke: @dkayiwa That feature was not in the specification https://wiki.openmrs.org/display/docs/Patient+Summary+Widget+Documentation#PatientSummaryWidgetDocumentation-conceptlistLatestObsForConceptList

dkayiwa: Does that mean the original code was some how buggy?

ssmusoke: @dkayiwa It was never documented in the spec that was being used to develop the widget

dkayiwa: Wouldn't removing it have effects on those who were depending on it?
Can't we improve the performance without altering existing behaviour?

ssmusoke: Let me add back the check, but I am wondering how they are using it when its not documented

dkayiwa: Am not sure. I do not even know if any one is using it. Am just cautious of introducing unintended effects in a ticket which is meant to simply improve performance.

ssmusoke: @dkayiwa  I think we should remove the check as its undocumented behavior, which is also not configurable as part of the widget app. Why should the widget limit the maxiumum number of concepts with no way to change it

dkayiwa: Am not against removing it. But am very much against doing so using a ticket whose only purpose is improving performance. This should not have any side effects other than that.

brandones: NB: This seems similar to https://issues.openmrs.org/browse/RA-1531, though that's about the ObsAcrossEncounters widget.

Maybe the right thing to do here is file a new ticket and attach it to this PR as "also fixes"?

dkayiwa: Having one commit to address two tickets is not a good practice.

omod/src/main/java/org/openmrs/module/coreapps/web/resource/LatestObsResource.java
mogoodrich: I think we've started just doing 2.* in many cases, not sure if this is "best practice" or not.

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
brandones: It's not uncommon for big overhauls, refactors, etc. to fix a handful of bugs along the way, just by replacing old code with a totally new and simpler approach. This kind of thing happens all the time, and is part of the point of a refactor.

You're right that it's a bit messy in terms of process. But I think it's worth it not to be intentionally re-implementing bugs.

dkayiwa: Even reviewing code that does more than what a ticket was meant for, becomes confusing to the reviewer. It has always been our practice that if there is a bug, one creates a new ticket of type bug, and then raises a pull request for it. The ticket includes a reasonable amount of documentation about the bug and why it should be fixed this way. Looking at this commit yourself, after a while, you can take some time before remembering how all these changes are related to the the performance improvement.

ssmusoke: I have updated the PR to include the feature limiting the maximum number of concepts on the widget to 10

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
```diff
@@ -15,39 +15,44 @@ export default class LatestObsForConceptListController {
...
-                            var date = this.$filter('date')(new Date(obs.value), this.config.dateFormat);
-                            obs.value = date;
-                        } else if (angular.isDefined(obs.value.display)) {
-                            //If value is a concept
-                            obs.value = obs.value.display;
-                        }
-
-                        this.obs.push(obs);
+        // from the number of elements specified by maxConceptCount remove the remaining elements
+        let elements_to_remove = concepts.length - this.maxConceptCount;
```
dkayiwa: I thought we had agreed to deal with maxConceptCount in a different pull request?

ssmusoke: @dkayiwa yes removing it, here I am maintaining the same functionality as it was before. 

Due to the change in the implementation even the restriction of concepts to a maximum had to change which was my earlier argument.

So this works as before limiting to only 10 concepts for the widget

PR #196, opened 2019-06-18
RA-1598: Allow for client-side filtering of diagnoses by concept sour…

Ruhanga:
…ces specified through a tag attribute

Ticket : https://issues.openmrs.org/browse/RA-1598


api/src/test/java/org/openmrs/module/coreapps/htmlformentry/EncounterDiagnosesTagHandlerTest.java
```diff
@@ -75,6 +85,46 @@ public void getSubstitution_shouldAddEncounterDiagnosisElementToFormSubmissionCo
...
+        attributes.put(CoreAppsConstants.HTMLFORMENTRY_ENCOUNTER_DIAGNOSES_TAG_INCLUDE_PRIOR_DIAGNOSES_ATTRIBUTE_NAME, "admit");
+        attributes.put("selectedDiagnosesTarget", "example-target");
+        attributes.put("conceptSource", "");
+        String hiddenInputElement = "\n<input type=\"hidden\" id=\"concept-source\" value=\"ICPC2\"/>\n";
+
+        // replay
+        String generatedHtml = encounterDiagnosesTagHandler.getSubstitution(formEntrySession, formSubmissionController, attributes);
+        
+        // verify
+        assertFalse(StringUtils.contains(generatedHtml, hiddenInputElement));
```
mks-d: If `generatedHtml` were to contain
```html
<input type="hidden" id="concept-source" value="ICD-10-WHO"/>
```
then this test would be misleading.

Maybe this would be "enough":
```java
String hiddenInputElement = "\n<input type=\"hidden\" id=\"concept-source\"";
```
In that case it should be renamed `hiddenInputElementStartingPart` or something like that.

A proper test would actually be to parse `generatedHtml` and ensure that there is no tag ID's `"concept-source", probably overkill.

mogoodrich:
@mks-d  I'm on-site in Malawi this week, will try to review but I might not get to it until next week... feel free to ping me again next week if I haven't responded!

Ruhanga:
Yes, @mks-d, this makes sense. I have pushed a commit to this effect.

api/src/main/resources/messages.properties
```diff
@@ -511,7 +511,7 @@ coreapps.consult.primaryDiagnosis=Primary Diagnosis:
 coreapps.consult.primaryDiagnosis.notChosen=Not chosen
 coreapps.consult.secondaryDiagnoses=Secondary Diagnoses:
 coreapps.consult.secondaryDiagnoses.notChosen=None
-coreapps.consult.codedButNoCode=ICD code unknown
+coreapps.consult.codedButNoCode=code unknown
```
mks-d: Wasn't it possible to do something like
```java
coreapps.consult.codedButNoCode={0} code unknown
```
?

omod/src/main/webapp/fragments/diagnosis/encounterDiagnoses.gsp
```diff
@@ -13,7 +13,7 @@
         {{ if (item.code) { }}
         {{- item.code }}
         {{ } else if (item.concept) { }}
-        ${ui.message("coreapps.consult.codedButNoCode")}
+        {{- item.preferredCodingSource }} ${ui.message("coreapps.consult.codedButNoCode")}
```
mks-d: And here something like (this has to be checked):
```gsp
${ui.message("coreapps.consult.codedButNoCode", item.preferredCodingSource)}
```

Ruhanga: This is quite impossible to do because `item.preferredCodingSource` is javascript that is referred to after `ui.message()` has done its part. Maybe such an usage as passing `preferredCodingSource` through the fragment's `config` to do the same such as 

`${ui.message("coreapps.consult.codedButNoCode", config.preferredCodingSource)}`

Just a thought. But let me check this out.

Ruhanga: So @mks-d, the above approach using `item.preferredCodingSource` fails with a [pre-mature error](https://hastebin.com/erudamebon.coffeescript). I've just made a branch of this one with working code [here](https://github.com/Ruhanga/openmrs-module-coreapps/pull/2/commits/9637440cce3a68d804f7a8a6f4e7b6df381b9382). So I should go ahead with this approach. And to make the solution uniform, I'd have to refactor the changes here so the `preferredCodingSource` is acually placed within `<input id="diagnosis-search . . . . ./>"` like was similarly done on RA-1603.

mks-d: If this works, let's settle on that:
```java
${ui.message("coreapps.consult.codedButNoCode", config.preferredCodingSource)}
```

omod/src/main/webapp/resources/scripts/diagnoses/diagnoses.js
```diff
@@ -1,10 +1,17 @@
 (function(diagnoses, $, _, undefined) {
-
+    
     var codingSystemToUse = 'ICD-10-WHO';
```
mogoodrich: we should now change this variable to 'defaultPreferredCodingSource'

omod/src/main/webapp/resources/scripts/diagnoses/diagnoses.js
```diff
@@ -1,10 +1,17 @@
...
 
     var mapTypeOrder = [ "SAME-AS", "NARROWER-THAN" ]
+    
+    findPreferredCodingSource = function(preferredCodingSource) {
```
mogoodrich: does this really need to take in a "preferreedConfingSource"... it makes it confusing... I whould thing this function shouldn't need to tak ein a parameter...

omod/src/main/webapp/resources/scripts/diagnoses/diagnoses.js
```diff
@@ -1,10 +1,17 @@
...
+        var codingSource = $("#diagnosis-search").attr("preferredcodingsource");
+        preferredCodingSource = codingSource != null ? codingSource : preferredCodingSource;
```
mogoodrich: ... and then this line would be...

```
var preferreedCodingSource = ? codiingSource != null ? codingSource: defaultPreferredCodingSource
```

Ruhanga:
Thank you @mogoodrich. I have addressed the requested changes. Let me know if this is okay to merge.

Ruhanga:
@mogoodrich, @mks-d, are we good to go?  :)

omod/src/main/webapp/resources/scripts/diagnoses/diagnoses.js
```diff
@@ -1,10 +1,17 @@
 (function(diagnoses, $, _, undefined) {
-
-    var codingSystemToUse = 'ICD-10-WHO';
+    
+    var defaultPreferredCodingSource = 'ICD-10-WHO';
```
mks-d: Why would this still be here if we introduced `CoreAppsConstants.DEFAULT_CODING_SOURCE`? That's not ok. Somehow this JS should be fed with the default preferred coding source coming from the backend.

omod/src/main/webapp/resources/scripts/diagnoses/diagnoses.js
```diff
@@ -1,10 +1,14 @@
 (function(diagnoses, $, _, undefined) {
 
-    var codingSystemToUse = 'ICD-10-WHO';
-
     var mapTypeOrder = [ "SAME-AS", "NARROWER-THAN" ]
+    
+    findPreferredCodingSource = function() {
+        var preferredCodingSource = $("#diagnosis-search").attr("preferredcodingsource");
+        return preferredCodingSource;
```
mogoodrich: What happened to the default preferred coding source?

I was asking to refactor this method so that it would return the default coding source (ICD-10-WHO) if nothing was specified as a attribute, instead of handling the default in the findConceptMapping method, not remove it entirely... or is this handled somewhere else now?  See my pseudo-code in the previous review.

mogoodrich:
@Ruhanga I added one follow-on comment...

omod/src/main/webapp/resources/scripts/diagnoses/diagnoses.js
Ruhanga: Thank you @mogoodrich, yes this is handled at the back end in the Element here https://github.com/openmrs/openmrs-module-coreapps/pull/196/files#diff-7d6c8132f77707ede46ad8ec6a618a59R80 . @mks-d highlighted on this on the previous comment.

omod/src/main/webapp/resources/scripts/diagnoses/diagnoses.js
Ruhanga: Sorry about my late response @mks-d, my comment could have disappeared in a draft. I addressed this in the last latest commit.

omod/src/main/webapp/resources/scripts/diagnoses/diagnoses.js
mogoodrich: Cool, thanks!

mogoodrich:
This okay by me if it looks good to @mks-d

PR #214, opened 2019-07-16
RA-1611 : Follow up commit for diagnosisSets and diagnosisClasses par…

Ruhanga:
…ameters alternate between 'null' and the 'default' from the their GPs

PR : https://issues.openmrs.org/browse/RA-1611

### DIFF
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/DiagnosesFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/DiagnosesFragmentController.java
@@ -58,9 +58,9 @@ public class DiagnosesFragmentController {
                                      @SpringBean("emrConceptService") EmrConceptService emrConceptService,
                                      @SpringBean("conceptService") ConceptService conceptService,
                                      @RequestParam("term") String query,
-                                     @RequestParam(value = "diagnosisSets", defaultValue = "") String diagnosisSetUuids,
-                                     @RequestParam(value = "diagnosisConceptSources", defaultValue = "") String diagnosisConceptSources,
-                                     @RequestParam(value = "diagnosisConceptClasses", defaultValue = "") String diagnosisConceptClasses,
+                                     @RequestParam(value = "diagnosisSets", required = false) String diagnosisSetUuids,
+                                     @RequestParam(value = "diagnosisConceptSources", required = false) String diagnosisConceptSources,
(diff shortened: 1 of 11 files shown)


omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/DiagnosesFragmentController.java
```diff
@@ -89,7 +89,7 @@
             }
         }
         else {
-            diagnosisSets = emrApiProperties.getDiagnosisSets();
+            diagnosisSets = StringUtils.isNotEmpty(diagnosisConceptSources) || StringUtils.isNotEmpty(diagnosisConceptClasses) ? null : emrApiProperties.getDiagnosisSets();
```
mks-d: We said:
>as soon as we start specifying any attribute, all defaults drop

For the sake of this controller, should the check not be
```java
diagnosisConceptSources != null || diagnosisConceptClasses != null
```
?

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/DiagnosesFragmentController.java
```diff
@@ -105,7 +105,7 @@
             }
         }
         else {
-            sources = emrApiProperties.getConceptSourcesForDiagnosisSearch();
+            sources = StringUtils.isNotEmpty(diagnosisSetUuids) || StringUtils.isNotEmpty(diagnosisConceptClasses) ? null : emrApiProperties.getConceptSourcesForDiagnosisSearch();
```
mks-d: For the sake of this controller, should the check not be
```java
diagnosisSetUuids != null || diagnosisConceptClasses != null
```
?

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/DiagnosesFragmentController.java
Ruhanga: Ohh, sorry I didn't clarify that actually spring places a null string/character in the requested parameter's variable. I attempted this earlier and that check failed then did a failover to apache `StringUtils`.

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/DiagnosesFragmentController.java
Ruhanga: The same story as above happens here @mks-d.

PR #243, opened 2019-10-19
RA-1633:Encounter type internationalization (Encounter Section on Patient dashboard)

mukesh-joomindia:
Current implementation does not support  encounter type name internationalization on patient dashboard in encounter section of visit tab(for captured encounter).

### DIFF
--- a/api/src/main/java/org/openmrs/module/coreapps/parser/ParseEncounterToJson.java
+++ b/api/src/main/java/org/openmrs/module/coreapps/parser/ParseEncounterToJson.java
@@ -45,7 +45,7 @@ public class ParseEncounterToJson {
 
 
         SimpleObject simpleEncounter = SimpleObject.fromObject(new EncounterDomainWrapper(encounter), uiUtils, "encounterId",
-                "location", "encounterDatetime", "encounterProviders.provider", "voided", "form","encounter.encounterType.name");
+                "location", "encounterDatetime", "encounterProviders.provider", "voided", "form");
 
         // UUID is not provided by EncounterDomainWrapper, adding it here.
         simpleEncounter.put("uuid", encounter.getUuid());
--- a/api/src/main/java/org/openmrs/module/coreapps/parser/ParseEncounterToJson.java
+++ b/api/src/main/java/org/openmrs/module/coreapps/parser/ParseEncounterToJson.java
@@ -45,7 +45,7 @@ public class ParseEncounterToJson {
 
 
         SimpleObject simpleEncounter = SimpleObject.fromObject(new EncounterDomainWrapper(encounter), uiUtils, "encounterId",
-                "location", "encounterDatetime", "encounterProviders.provider", "voided", "form");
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/coreapps/parser/ParseEncounterToJson.java
```diff
@@ -45,7 +45,7 @@ public SimpleObject createEncounterJSON(User authenticatedUser, Encounter encoun
 
 
         SimpleObject simpleEncounter = SimpleObject.fromObject(new EncounterDomainWrapper(encounter), uiUtils, "encounterId",
-                "location", "encounterDatetime", "encounterProviders.provider", "voided", "form");
+                "location", "encounterDatetime", "encounterProviders.provider", "voided", "form","encounter.encounterType.name");
```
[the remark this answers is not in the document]
mukesh-joomindia: There is no need of extra field ("encounter.encounterType.name") in model.I removed this field.

api/src/main/java/org/openmrs/module/coreapps/parser/ParseEncounterToJson.java
```diff
@@ -68,7 +68,7 @@ public SimpleObject createEncounterJSON(User authenticatedUser, Encounter encoun
         }
         EncounterType encounterType = encounter.getEncounterType();
         simpleEncounter.put("encounterType",
-                SimpleObject.create("uuid", encounterType.getUuid(), "name", uiUtils.format(encounterType)));
+                SimpleObject.create("uuid", encounterType.getUuid(), "name", uiUtils.message(encounterType.getName())));
```
mks-d: Before your changes the entire encounter type representation was passed to the model. Let's try to not break this, could you try this:
```java
EncounterType encounterType = encounter.getEncounterType();
encounterType.setName(uiUtils.message(encounterType.getName()));
simpleEncounter.put("encounterType",
        SimpleObject.create("uuid", encounterType.getUuid(), "name", uiUtils.format(encounterType)));
```

mukesh-joomindia: Yes ,we can.If you see https://github.com/openmrs/openmrs-module-uiframework/blob/fcc39f0b7e4cccb04834d10d1c5d6e1dec29a523/api/src/main/java/org/openmrs/ui/framework/FormatterImpl.java#L65-L96 , uiutils.format() checks a lot of class and that is overburden ,we already know that at this point we need only encounter type name so instead of using uiutils.format()  ,uiUtils.message() would be better.

mks-d: @mogoodrich do you know if anything else than the encounter type name is used  on those views?

mogoodrich: Whoops, sorry I didn't catch this earlier, there's already a way to do this, will comment on the ticket.

PR #253, opened 2019-11-19
RA-1638: Renaming widget "appointments" to "bamniappointments".

mozzy11:
Ticket : https://issues.openmrs.org/browse/RA-1638 
Follow up commit to rename the bahmni appointments widget from "appointments"  to "bahmniappointments"


mozzy11:
cc @mks-d

PR #256, opened 2019-12-05
RA-1641: inactive conditions should be displayed

mozzy11:
Ticket: https://issues.openmrs.org/browse/RA-1641
Modified End point to fetch all Conditions both active and inactive


mozzy11:
cc @mseaton

PR #257, opened 2019-12-11
RA-1653: Program Status widgets, single call to location REST endpoint.

Ruhanga:
Ticket : https://issues.openmrs.org/browse/RA-1653

### DIFF
--- a/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
+++ b/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
@@ -197,12 +197,20 @@ export default class ProgramStatusController {
     }
 
     fetchLocations() {
-        return this.openmrsRest.get('location', {
-            v: 'custom:display,uuid',
-            tag: this.config.locationTag,
-        }).then((response) => {
-            this.programLocations = response.results;
-        })
+        var locationsPromise = this.$window.sessionContext.locationsPromise
+        if(!locationsPromise) {
+            locationsPromise = this.openmrsRest.get("location", {
+                v: "custom:display,uuid",
+                tag: this.config.locationTag
+            })
+            // make the location promise available for other program widgets
+            // sessionContext is defined in appui - https://github.com/openmrs/openmrs-module-appui/blob/master/omod/src/main/webapp/fragments/decorator/standardEmrPage.gsp
+            this.$window.sessionContext.locationsPromise = locationsPromise
+        }
+        return locationsPromise.then(e => {
+            this.programLocations = e.results
+            return e
+        });
     }
 
     fetchSessionLocation() {


omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
```diff
@@ -197,12 +197,19 @@ export default class ProgramStatusController {
...
-            this.programLocations = response.results;
-        })
+        var locationsPromise = this.$window.sessionContext.locationsPromise
+        if(!locationsPromise) {
+            locationsPromise = this.openmrsRest.get("location", {
+                v: "custom:display,uuid",
+                tag: this.config.locationTag
+            })
+            // make the location promise available for other program widgets
+            this.$window.sessionContext.locationsPromise = locationsPromise
```
Ruhanga: Hi @mogoodrich, I ended up using the `sessionContext` object to contain the the locations promise so all the existing program status widgets can use it once to populate their locations. Do you anticipate any issue with this approach? Thanks.

mogoodrich: Thanks @Ruhanga .  I'm not familiar with window.sessionContext... can you point me to do some documentation of it?  Another option would be to use window.sessionStorage... https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage

Ruhanga: Ahh, great. Thanks @mogoodrich. `sessionContext` is defined [in appui](https://github.com/openmrs/openmrs-module-appui/blob/master/omod/src/main/webapp/fragments/decorator/standardEmrPage.gsp#L28) and is used mainly for the locale info. I'm going to use this `window.sessionStorage` object instead. Thank you :-)

mogoodrich: Oh interesting!  It doesn't look like the "window.sessionContext" is used often (or at all) so sessionStorage is probably a better bet.

Ruhanga: @mogoodrich, so it is impossible to store objects with functions in the `sessionStorage`. `sessionStorage` does not handle events. This makes it difficult to store a promise or use some event handling mechanism that updates multiple program status widgets with the fetched locations. Perhaps I may create another `window.` say `programEnrollmentLocationsPromise`.  `programEnrollmentLocationsPromise` is quite verbose though. :(

mogoodrich: @Ruhanga ah... feel free to use sessionContext then, I'd just add a comment to the code explaining where/how that is defined.

Ruhanga:
Great. Thanks @mogoodrich.

PR #260, opened 2019-12-19
RA-1665: 'latestObsForConceptList' dashboard widget to support obs groups at first level.

Ruhanga:
Ticket : https://issues.openmrs.org/browse/RA-1665


omod/src/main/java/org/openmrs/module/coreapps/web/resource/LatestObsResource.java
```diff
@@ -85,9 +86,12 @@ protected PageableResult doSearch(RequestContext context) {
 				List<Concept> questions = new Vector<Concept>();
 				questions.add(concept);
 				
-				List<Obs> obs = obsService.getObservations(who, null, questions, null, null, null, sort, 1, null, null, null, false);
-				if (obs.size() > 0) {
+				int nLatestObsPerGroup = StringUtils.isNotEmpty(context.getParameter("nLatestObsPerGroup")) ? Integer.parseInt(context.getParameter("nLatestObsPerGroup")) : 1;
```
mogoodrich: Maybe just call these "nLatestObs"?  I think "perGroup" is a little confusing because there's only one "group" here, the group of all the obs associated wtih the list of question concepts, correct?  I also think using "group" is confusing because an "ObsGroup" has a specific meaning and I don't think you are using it in that context if I'm understanding correct?

Will add some more comments on the  ticket.

Ruhanga: Oh yes, thanks @mogoodrich. `perGroup` makes it verbose and confusing. I'll correct this in the next commit.

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
```diff
@@ -38,8 +38,33 @@ export default class LatestObsForConceptListController {
...
                 if (angular.isUndefined(this.maxAgeInDays) || this.widgetsCommons.dateToDaysAgo(obs.obsDatetime) <= this.maxAgeInDays) {
                     // Add last obs for concept to list
-
-                    if (['8d4a505e-c2cc-11de-8d13-0010c6dffd0f',
+                	
+                	if (angular.isDefined(obs.groupMembers) && obs.groupMembers != null) {
+                        // If obs is obs group
+                        obs.members = [];
+                        obs.value = "";
+                        angular.forEach(obs.groupMembers, (member) => {
```
mks-d: I find it confusing to have both an `obs.groupMembers` and an `obs.members` on top of/alongside it. Why not just _transform_ `obs.groupMembers` to be the array that you want?
Also it would be better practise to evaluate `config.obsGroupLabels` outside of the `forEach`. This is the kind of stuff that eventually backfires, performance wise.

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
```diff
@@ -38,8 +38,33 @@ export default class LatestObsForConceptListController {
...
+                        		prefix = "(" + member.concept.display + " #" + obs.value.split(member.concept.display).length + ") ";
+                        	} else if (angular.isDefined(this.config.obsGroupLabels) && this.config.obsGroupLabels == "shortName") {
+                        		prefix = "(" + member.concept.name.display + " #" + obs.value.split(member.concept.name.display).length + ") ";
+                        	} else {
+                                // for default  or obsGroupLabels = none option
+                                prefix = "";
+                        	}
+                        	
+                            // this will help determine the count of obs with same concept
+                            if (obs.value === "") {
```
mks-d: Ah ok, we don't need this whole logic here. The `#1`, `#2`, `#3`, ... etc were supposed to be integral part of the FSNs or short names, I used those counts to illustrate the case of an obs group made of members defined by separate/different concepts. I didn't mean that they should be counted on the display. That should simplify this whole code quite a bit.

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
Ruhanga: Thanks @mks-d. Yes, I'll simply transform `obs.groupMembers`. In the case of `forEach`, I would have to replicate its code three times or else creating a generic method still adds more `if, else-if` checks, so I guess we would leave the `forEach` as is.

Ruhanga:
Ahh, nice catch @mogoodrich. Thank you. Let me address this one.

PR #262, opened 2020-01-07
RA-1641:Condition list--inactive conditions are not displayed

gitcliff:
LINK:https://issues.openmrs.org/browse/RA-1641
Condition list--inactive conditions are not displayed


mozzy11:
Looks good

mozzy11:
@dkayiwa , i tested this and it works fine

mogoodrich:
Are we sure that the condition history is always sorted properly from earliest to latest? What happens if I add a condition and then mark it inactive on the same day?

mozzy11:
@mogoodrich  sure its sorted poperly from earliest to latest. I did several testing regarding that. Actually all the testing i did , i was adding several conditions , activating , and de-activating them on the same day. The condition History is properly sorted. The order in which the condition history is persisted in the database , is the same order in which it is retrieved.

mogoodrich:
Thanks @mozzy11 I will do some testing!

PR #268, opened 2020-01-14
Program Status widget: format using bootstrap form-control and form-c…

mogoodrich:
…ontrol-sm

### DIFF
--- a/omod/src/main/web/dashboardwidgets/datepicker/datepicker.html
+++ b/omod/src/main/web/dashboardwidgets/datepicker/datepicker.html
@@ -1,5 +1,5 @@
 <div class="input-group date">
-    <!-- TODO: make form-control-sm configurable -->
+    <!-- TODO: make whether or not form-control-sm class is applied to the input configurable -->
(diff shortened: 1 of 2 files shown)


omod/src/main/web/dashboardwidgets/datepicker/datepicker.html
```diff
@@ -1,5 +1,6 @@
 <div class="input-group date">
-    <input type="text" class="form-control" readonly="readonly">
+    <!-- TODO: make form-control-sm configurable -->
```
brandones: What do you mean by this? You shouldn't override bootstrap classes. Just add additional classes when needed.

mogoodrich: I just meant that users of the openmrs-datepicker widget should be able to choose between the bootstrap normal and bootstrap small layout (ie whether or not the form-control-sm class is applied here).  I'll update the comment, to (hopefully!) make it clearer.

PR #274, opened 2020-01-22
RA-1628:Void conditions history

gitcliff:
https://issues.openmrs.org/browse/RA-1677

### DIFF
--- a/omod/src/main/webapp/resources/scripts/conditionlist/controllers/manageconditions.controller.js
+++ b/omod/src/main/webapp/resources/scripts/conditionlist/controllers/manageconditions.controller.js
@@ -40,8 +40,16 @@ function ManageConditionsController($scope, RestfulService, CommonFunctions) {
         }
 
     self.removeCondition = self.removeCondition || function () {
-            $scope.conditionToBeRemoved.voided = true;
-            $scope.conditionToBeRemoved.endDate = new Date();
+            
+var conditions = [];
+var condition;
+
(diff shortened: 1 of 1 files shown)


omod/src/main/webapp/resources/scripts/conditionlist/controllers/manageconditions.controller.js
```diff
@@ -40,8 +40,16 @@ function ManageConditionsController($scope, RestfulService, CommonFunctions) {
         }
 
     self.removeCondition = self.removeCondition || function () {
-            $scope.conditionToBeRemoved.voided = true;
-            $scope.conditionToBeRemoved.endDate = new Date();
+            
+var conditions = [];
```
mozzy11: this condition List is empty , it doesnt point to any thing

omod/src/main/webapp/resources/scripts/conditionlist/controllers/manageconditions.controller.js
```diff
@@ -40,8 +40,16 @@ function ManageConditionsController($scope, RestfulService, CommonFunctions) {
...
+var condition;
+
+      for(condition in conditions){
+       condition.voided = true;
+        condition.endDate = new Date();
+        console.log(condition);
```
mozzy11: this will actualy log nothing, your Condiitons List is , empty , it points to nothing

PR #276, opened 2020-01-24
RA-1628:Condition List:Voiding/Deleting condition should delete entire history

gitcliff:
link:https://issues.openmrs.org/browse/RA-1677


mozzy11:
LGTM . I have tested and it works  
Thanks @gitcliff 
cc @mogoodrich

PR #283, opened 2020-02-04
RA-1697: Improve Actions Column styling on the Patient DashBord

mozzy11:
Ticket :https://issues.openmrs.org/browse/RA-1697
Improve "Actions" Column styling on the Patient DashBord


mozzy11:
@dkayiwa  @mogoodrich  , we also need this commit  , so that we can release these remaining modules

PR #288, opened 2020-02-06
RA-1700, configure Visits Section widget

cioan:


cioan:
I think that AppDescriptor app = (AppDescriptor) pageModel.get("app") refers to the CLINICIAN_DASHBOARD app, the host page, and @FragmentParam("app") AppDescriptor appDescriptor refers to the Visits Section widget that is placed on the CLINICIAN_DASHBOARD page.

mogoodrich:
Got it @cioan ... can you confirm that and then add some comments referring to that so for the future since it's confusing? Otherwise, looks good!

cioan:
Thanks @mogoodrich !

PR #292, opened 2020-02-15
RA-1703, display concept SHORT names if available

cioan:
When obs values are coded answers, try first to display the SHORT name of the concept in the locale of the current session.

### DIFF
--- a/omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
+++ b/omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
@@ -7,6 +7,9 @@ export default class ObsAcrossEncountersController {
 
   $onInit() {
     this.order = 'desc';
+    this.sessionLocale = null;
+    this.CONCEPT_CUSTOM_REP = 'custom:(uuid,display,names:(display,conceptNameType,locale)';
+
     this.concepts = [];
     // a map of conceptUUID --> concept(REST response)
     this.conceptsMap = {};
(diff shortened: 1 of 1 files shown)


omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
```diff
@@ -98,11 +111,20 @@ export default class ObsAcrossEncountersController {
   }
 
   getConceptWithShortName(concept) {
-    angular.forEach(concept.names, (name) => {
-      if (name.conceptNameType == 'SHORT') {
-        concept.display = name.display;
+    let lastShortName = null;
+    for(let i=0; i < concept.names.length; i++) {
+      let conceptName = concept.names[i];
+      if (conceptName.conceptNameType == 'SHORT') {
```
CaptainDaVinci: I'm not familiar with the code base, but why not use '===' for a stricter comparison?

omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
```diff
@@ -145,9 +167,41 @@ export default class ObsAcrossEncountersController {
         };
         this.simpleEncs.push(tempEnc);
       }
+      this.updateWithConceptShortNames(this.simpleEncs);
     });
   }
 
+  updateWithConceptShortNames(encounters) {
```
mogoodrich: Let's add a comment here, "iterates through all obs, and for all value coded, replace the default obs.value.display  with the value display of the appropriate shortName"

PR #296, opened 2020-02-21
RA-1703, if specified via config property display concept short name

cioan:
Hi @mogoodrich , here is an updated PR for displaying concept SHORT names if configured via **_useConceptShortName_** widget config property. By default, it will continue to display the obs.value.display property. Thanks.


cioan:
Thanks @mogoodrich !

PR #297, opened 2020-02-28
RA-1709, check for errors when fetching user privileges

cioan:
@mogoodrich, @mseaton  , I just added some error handling when fetching the privileges. Hope this would provide more debug info when it fails next time.

### DIFF
--- a/omod/src/main/web/dashboardwidgets/relationships/relationships.controller.js
+++ b/omod/src/main/web/dashboardwidgets/relationships/relationships.controller.js
@@ -52,9 +52,11 @@ export default class RelationshipsController  {
 
         //fetchPrivileges
         this.openmrsRest.get('session', {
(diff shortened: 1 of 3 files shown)


omod/src/main/web/dashboardwidgets/relationships/relationships.controller.js
```diff
@@ -52,9 +52,11 @@ export default class RelationshipsController  {
 
         //fetchPrivileges
         this.openmrsRest.get('session', {
-            v: 'custom:(privileges)'
+            v: 'custom:(privileges:(name))'
```
mogoodrich: We've confirmed that it's okay to narrow down this REST call to only return the "name" of the privileges?  (I don't know why that wouldn't be fine, but just double-checking).

cioan:
Thanks @mogoodrich !

PR #299, opened 2020-03-06
RA-1704: Improved Program Status widget state sorting.

kml27:
JIRA Ticket: https://issues.openmrs.org/browse/RA-1704

Provides new program workflow state sort logic for ProgramStatus/Enrollment widget.

Observed:
Previously, states were only sorted by start date. This resulted in randomly incorrect order of states when the states were started on the same day, but one ended after the other and also when multiple states were started and ended on the same day.

Expected:
States displayed in the order the patient was entered into them.

Fix:
Sort by start date, those with the same start date then sorted by end date, those with the same end date then sorted by created date.

This also provides a jest/karma/angular/phantomjs test to prevent regression.

Additional changes:

 * updating angular to 1.6.1 to get $httpBackend.flush() fix (related to https://github.com/angular/angular.js/issues/14631 )
 * changes to use a proper promise chain for synchronous resolution during test promise chain and $httpBackend.flush()

### DIFF
--- a/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
+++ b/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
@@ -468,25 +468,37 @@ export default class ProgramStatusController {
                     //if index of state1 is 7 and state2 is 8, 7-8 == -1 indicates state1 is first, 8-7 == 1 indicating state2 is first
                     return state1.value-state2.value;
                 }
-
-                //if the end date is not set, or end date is after, the state occurred before
-                //if the start date of 1 is earlier, is occurs before 
-                //if both the end dates are identical, check the created at time
                 
                 //this sort is done against ALL workflows within a program 
                 //in which case it is possible to have two states with null end dates,
                 //and that is the exception to when endDate==null indicates sort order clearly
-                if((!(state1.value.endDate == null && state2.value.endDate == null) && state2.value.endDate == null) || state1.value.startDate < state2.value.startDate || state1.value.endDate < state2.value.endDate || state1.value.dateCreated < state2.value.dateCreated) {
-                    return -1;
-                    
-
-                //later if startdate occurs after
-                //later if the end date is later or both end dates are not null, but the first enddate is null
-                //if both the end dates are identical, check the created at time
-                } else if ((!(state1.value.endDate == null && state2.value.endDate == null) && state1.value.endDate == null) || state1.value.startDate > state2.value.startDate || state1.value.endDate > state2.value.endDate || state1.value.dateCreated > state2.value.dateCreated) {
-                    return 1;    
-                }
+                if(!(state1.value.endDate == null && state2.value.endDate == null)) {
                 
+                    //ordered so each criteria is evaluated in order (e.g. not all "prev" crit. before "next" crit. is ever evaluated)
+
+                    //null end date prioritized over any other sort criteria
+                    if(state2.value.endDate == null){
+                        return -1;
+                    } else if( state1.value.endDate == null){
+                        return 1;
+                    //sort by start date
+                    } else if( state1.value.startDate < state2.value.startDate) {
+                        return -1;
+                    } else if ( state1.value.startDate > state2.value.startDate) {
+                        return 1;    
+                    //sort by end date
+                    } else if(state1.value.endDate < state2.value.endDate){
+                        return -1;
+                    } else if (state1.value.endDate > state2.value.endDate) {
+                        return 1;
+                    //sort by date created
+                    } else if(state1.value.dateCreated < state2.value.dateCreated){
+                        return -1;
+                    } else if(state1.value.dateCreated > state2.value.dateCreated){
+                        return 1;
+                    }
+
+                }
             	return 0;
             });
 
--- a/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
+++ b/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
@@ -461,9 +461,18 @@ export default class ProgramStatusController {
             this.patientProgram.states = this.$filter('filter')(this.patientProgram.states, (state) => {
                 return !state.voided
             }, true);
+            
+            //this custom comparator is detailed in the PR at https://github.com/openmrs/openmrs-module-coreapps/pull/299
+            //the orderBy docs are available at https://docs.angularjs.org/api/ng/filter/orderBy
             this.patientProgram.states = this.$filter('orderBy')(this.patientProgram.states, null, false, function(state1, state2)
             {
-                //orderBy falls back to index comparison when two states are indicated to be equal
+                //From the orderBy documentation
+                
+                //In order to ensure that the sorting will be deterministic across platforms, if none of the specified predicates can
+                //distinguish between two items, orderBy will automatically introduce a dummy predicate that returns the item's index
+                //as value. (If you are using a custom comparator, make sure it can handle this predicate as well.)
+                
+                //i.e. orderBy falls back to index comparison when two states are indicated to be equal (return 0) in a previous comparison
                 if(state1.type === "number" && state2.type === "number"){
                     //if index of state1 is 7 and state2 is 8, 7-8 == -1 indicates state1 is first, 8-7 == 1 indicating state2 is first
                     return state1.value-state2.value;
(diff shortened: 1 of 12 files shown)


omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
```diff
@@ -466,8 +463,42 @@ export default class ProgramStatusController {
             this.patientProgram.states = this.$filter('filter')(this.patientProgram.states, (state) => {
                 return !state.voided
             }, true);
-            this.patientProgram.states = this.$filter('orderBy')(this.patientProgram.states, (state) => {
-                return new Date(state.startDate);
+            this.patientProgram.states = this.$filter('orderBy')(this.patientProgram.states, null, false, function(state1, state2)
+             {
+            	var result = 0;
+            	
+            	//console.log(state1, state2);
```
mogoodrich: let's remove this commented out log steatement (and all the ones below as well)

omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
```diff
@@ -466,8 +463,42 @@ export default class ProgramStatusController {
...
+            	
+            	//if the start date of 1 is earlier, is occurs before 
+            	if(state1.value.startDate < state2.value.startDate) {
+            		//console.log("based on startDate", state1.value.startDate, "is before", state2.value.startDate);
+            		result = -1;
```
mogoodrich: others may have a different preference, but I'd prefer just a return -1 here...

omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
```diff
@@ -466,8 +463,42 @@ export default class ProgramStatusController {
...
+            	//later it occurs after
+            	} else if (state1.value.startDate > state2.value.startDate) {
+            		result = 1;
```
mogoodrich: and a return 1 here... just let anyone following the code realize they don't need to read through further for this path.

omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
```diff
@@ -466,8 +463,42 @@ export default class ProgramStatusController {
...
+            	} else {
+            	
+	            	//if the end date is not set, or end date is after, the state occurred before
+	            	if(state1.value.endDate < state2.value.endDate || state2.value.endDate == null ) {
```
mogoodrich: Is this null-safe?  Should the null checks happen first?

kml27: I'll switch it to the null check first. `Date() < null`  and `Date() > null` are both `false` in Chrome, but I don't know if that behavior is specifically required by js/EMCAScript.

omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
```diff
@@ -466,8 +463,33 @@ export default class ProgramStatusController {
...
             }, true);
-            this.patientProgram.states = this.$filter('orderBy')(this.patientProgram.states, (state) => {
-                return new Date(state.startDate);
+            this.patientProgram.states = this.$filter('orderBy')(this.patientProgram.states, null, false, function(state1, state2)
+            {
+                //orderBy falls back to index comparison when two states are indicated to be equal
+                if(state1.type === "number" && state2.type === "number"){
+                    //if index of state1 is 7 and state2 is 8, 7-8 == -1 indicates state1 is first, 8-7 == 1 indicating state2 is first
+                    return state1.value-state2.value;
+                }
```
mogoodrich: Why did we end up adding a comparsion by index here?  Are we relying on the service returning the states ordered by DB primary key (when all else is equal)?  We should never be relying on index because although in the general use case earlier created states have lower primary keys, we make no guarantees, for example in term of migrations, synced databases, etc.

kml27: This is actually an angular behavior (and relates to the index of the element in the array being sorted)
```
In order to ensure that the sorting will be deterministic across platforms, if none of the specified predicates can distinguish between two items, orderBy will automatically introduce a dummy predicate that returns the item's index as value. (If you are using a custom comparator, make sure it can handle this predicate as well.)
```
- https://docs.angularjs.org/api/ng/filter/orderBy

omod/src/main/web/dashboardwidgets/relationships/relationships.controller.js
```diff
@@ -51,9 +51,7 @@ export default class RelationshipsController  {
         }
 
         //fetchPrivileges
-        this.openmrsRest.get('session', {
-            v: 'custom:(privileges:(name))'
-        }).then((response) => {
+        this.openmrsRest.get('session').then((response) => {
             this.getPrivileges(response.user.privileges);
```
kml27: Did this merge go wrong?

kml27: It looks like master... guess it's an old diff?

mogoodrich: This is a recent change we made to fix this: https://issues.openmrs.org/browse/RA-1719

kml27: Thanks, good to know. I see some privilege related differences in the program status widgets in the distro I'm working on (none of the edit/delete et c features are shown). Should we update to the latest RESTWS? @mks-d

mogoodrich: yes, if you are using the latest snapshot of core apps, you'll need the latest snapshot of REST for this to still work... I do plan to release a new version of REST WS with the fix in the coming days.

omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
mogoodrich: Wow, interesting... still not sure I 100% understand, though. :)

Can we extract this into an util method and through some test cases around it?

kml27: Which part? The orderby comparator? 

>  still not sure I 100% understand, though. :)

I was seeing it in the test. There are two program workflows, so it's possible to have two null end dates. Firefox was tending to sort two states in reverse order vs. chrome.

If the comparator says the two states are equal ( return 0; ), the two states are re-evaluated with the same comparator by their array index to provide a stable sort (the order is always the same given the same "verbatim" input, vs. an unstable sort where any two "equal" states could be in either order given that same input, generally this is seen as a symptom of quick sort pivot randomness).

mogoodrich: Ah, I thought you had added test cases, but then I didn't see them when I last checked because Github had collapsed that file because it was large... I'll check out the changes now.

mogoodrich: What confuses me about this is the fact that the fallback test appears first... it is that when orderBy calls the custom comparator and the comparator returns 0 it then converts the value to a number based on index and calls the comparator again?

kml27: > it is that when orderBy calls the custom comparator and the comparator returns 0 it then converts the value to a number based on index and calls the comparator again?

Yes (pretty much). It doesn't "convert to a number" so much as just fallback to passing the index instead of the obj. If you're not familiar with JS array prototypes, most (all?), like Array.prototype.forEach, can optionally receive the index, 
```
[3,2,1].forEach( function(obj, index) { console.log("array[",index,"]=",obj) } );

array[ 0 ]= 3
array[ 1 ]= 2
array[ 2 ]= 1
```

> What confuses me about this is the fact that the fallback test appears first...

I added it first as the "base case" so the rest of the code after it doesn't have to check for stateX.type="number" or vice versa. This is essentially the "type checking" code.

JS is weakly typed, and Angular converts this into it's own object as described in the [orderby doc](https://docs.angularjs.org/api/ng/filter/orderBy). Even in Java it wouldn't be as simple as just having two overloaded function signatures on a comparator (though that would be nice), because the ABC requires the Object function signature be overriden and implemented (which is where at least the type checking would occur, but could call corresponding functions for a given type instead of doing all the comparisons in the one function).

If you prefer these "two" comparator logic blocks (one for "number" and one for the state "object") be made into functions and called based on the "type checking" for clarity, I can accommodate.

mogoodrich: No, that's fine, works for me!  thanks for the detailed description!  Perhaps just add a comment to the text either linking to this discussion or some other explanation would be helpful, because I think others may be confused like me.

kml27: Ok, Thanks!

PR #300, opened 2020-03-09
RA-1719: Program, Program Status and Relationship widgets not properl…

mogoodrich:
…y retrieving session use

### DIFF
--- a/omod/src/main/web/dashboardwidgets/programs/programs.controller.js
+++ b/omod/src/main/web/dashboardwidgets/programs/programs.controller.js
@@ -55,9 +55,7 @@ export default class ProgramsController {
     }
 
     fetchPrivileges() {
-        this.openmrsRest.get('session', {
-            v: 'custom:(privileges:(name))'
-        }).then((response) => {
+        this.openmrsRest.get('session').then((response) => {
             if (response && response.user && angular.isArray(response.user.privileges)) {
                 if (response.user.privileges.some( (p) => { return p.name === 'Task: coreapps.enrollInProgram'; })) {
                     this.canEnrollInProgram = true;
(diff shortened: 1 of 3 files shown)


rrameshbtech:
@mogoodrich The new change in the session url (removing custom) has effects in the response returned by the server. 

For example `https://demo.openmrs.com/openmrs/ws/rest/v1/session` returns below which does not have name for privileges (older one had). which is what checked in all the functions to validate whether the logged user has particular privilege or not.

```
 "privileges": [
            {
                "uuid": "d05118c6-2490-4d78-a41a-390e3596a237",
                "display": "Get Concept Datatypes",
                "links": [
                    {
                        "rel": "self",
                        "uri": "http://uat.openmrs-icrc.com/openmrs/ws/rest/v1/privilege/d05118c6-2490-4d78-a41a-390e3596a237"
                    }
                ]
            },
            {
                "uuid": "d05118c6-2490-4d78-a41a-390e3596a238",
                "display": "Get Concept Classes",
                "links": [
                    {
                        "rel": "self",
                        "uri": "http://uat.openmrs-icrc.com/openmrs/ws/rest/v1/privilege/d05118c6-2490-4d78-a41a-390e3596a238"
                    }
                ]
            },
```

Could you help us to fix this?

Ruhanga:
Thanks for this observation @rrameshbtech. This in a way is fixed by depending on version 2.28.0 and above of `webservices.rest` module.

cc: @mogoodrich.

mogoodrich:
Yes, sorry @rrameshbtech  and @Ruhanga ... this bug fix was across both REST web services and Core Apps and so Core Apps needs the late REST module... I did a release of REST, but neglected to update the dependency in Core Apps, my apologies... I just did that now...

mogoodrich:
Of course it looks like the build failed after my change... I'm in a meeting now but will check when I'm out.

mogoodrich:
For the record, I had issues updating the Core Apps module to use the latest REST, but I did release the REST module and the ref app now relies on that version.

PR #301, opened 2020-03-10
RA-1724: Fix NPE/undefined error in obsgraph dashboard widget.

Ruhanga:
Ticket : https://issues.openmrs.org/browse/RA-1724

### DIFF
--- a/omod/src/main/web/dashboardwidgets/obsgraph/obsgraph.controller.js
+++ b/omod/src/main/web/dashboardwidgets/obsgraph/obsgraph.controller.js
@@ -127,7 +127,7 @@ export default class ObsGraphController {
                 if (concept.legend === true) {
                   if (concept.uuid && concept.type === "obs") {
                     let obj = this.dataset.find(element => element.uuid === concept.uuid);
-                      if (angular.isDefined(obj) && angular.isDefined(obj.values) && obj.values !== null && (Object.keys(obj.values)).length > 0) {
+                      if (angular.isDefined(obj) && obj !== nul && angular.isDefined(obj.values) && obj.values !== null && (Object.keys(obj.values)).length > 0) {
                         // if this concept has any obs values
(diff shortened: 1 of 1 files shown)


omod/src/main/web/dashboardwidgets/obsgraph/obsgraph.controller.js
```diff
@@ -127,7 +127,7 @@ export default class ObsGraphController {
                 if (concept.legend === true) {
                   if (concept.uuid && concept.type === "obs") {
                     let obj = this.dataset.find(element => element.uuid === concept.uuid);
-                      if (angular.isDefined(obj.values) && obj.values !== null && (Object.keys(obj.values)).length > 0) {
+                      if (angular.isDefined(obj) && angular.isDefined(obj.values) && obj.values !== null && (Object.keys(obj.values)).length > 0) {
```
Ruhanga: @mogoodrich, I think also that `dataset.find(..)` might never return null and so no need to check for this.

mogoodrich: I guess what I meant was I could see the need for adding this check to avoid a potential NPE, but it didn't seem to coincide with the PR description "obsgraph dashboard widget should not require all conceptId(s)"... if this is just a follow-on comment to fix a NPE/undefined error, makes sense to me!

Ruhanga: Thanks @mogoodrich. The PR description is somewhat describing the undesired widget's behaviour that ends up being fixed by fixing the underlying `undefined/null` error which is not visible to the user. Makes sense though to redefine the the PR description to match exactly what is being done.

mogoodrich: No worries... thanks!

PR #306, opened 2020-03-18
RA-1729: Copying 'Dispensed Medication' widget from Order Entry UI.

Ruhanga:
Ticket : https://issues.openmrs.org/browse/RA-1729

### DIFF
--- a/api/src/main/resources/messages.properties
+++ b/api/src/main/resources/messages.properties
@@ -31,6 +31,7 @@ coreapps.age.days={0} Day(s)
 coreapps.patientDashBoard.editPatientIdentifier.duplicateMessage=is already in use
 coreapps.patientDashBoard.editPatientIdentifier.invalidMessage=did not pass validation
 coreapps.patientDashBoard.editPatientIdentifier.invalidFormatMessage=must be in the format
+coreapps.patientdashboard.activeDrugOrders=Dispensed Medication
 
 coreapps.startDate.label=Start Date
 coreapps.stopDate.label=End Date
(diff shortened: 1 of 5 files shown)


api/src/main/resources/messages.properties
```diff
@@ -31,6 +31,7 @@ coreapps.age.days={0} Day(s)
 coreapps.patientDashBoard.editPatientIdentifier.duplicateMessage=is already in use
 coreapps.patientDashBoard.editPatientIdentifier.invalidMessage=did not pass validation
 coreapps.patientDashBoard.editPatientIdentifier.invalidFormatMessage=must be in the format
+coreapps.patientdashboard.activeDrugOrders=Dispensed Medication
```
mks-d: Not sure if this should be plural.

omod/src/test/resources/dispensedMedication_app.json
```diff
@@ -0,0 +1,21 @@
+{
+  "id": "coreapps.dispensedMedication",
+  "instanceOf": "coreapps.template.dashboardWidget",
+  "description": "Show Dispenced Medication as drug orders",
```
mks-d:
```json
  "description": "Show dispensed medications as drug orders",
```

api/src/main/resources/messages.properties
Ruhanga: @mks-d, are you referring to something like `Dispensed Medications`?

omod/src/test/resources/dispensedMedication_app.json
Ruhanga: Ok 👍

api/src/main/resources/messages.properties
mks-d: Yes but no ;-) This word stays singular in English. We keep it as it is.

PR #307, opened 2020-03-18
RA-1715-Diagnosis not being showed in the visit summary column

reagan-meant:
https://issues.openmrs.org/browse/RA-1715
Am having an error which I need help with...Background information https://talk.openmrs.org/t/gsp-errors-related-to-commit/27466
@ibacher @dkayiwa @mks-d @mogoodrich @Ruhanga @mozzy11

### DIFF
--- a/omod/src/main/webapp/fragments/patientdashboard/visits.gsp
+++ b/omod/src/main/webapp/fragments/patientdashboard/visits.gsp
@@ -70,8 +70,15 @@
     <ul id="visits-list" class="left-menu">
         <%
             def visits = patient.allVisitsUsingWrappers;
+
             visits.eachWithIndex { wrapper, idx ->
-                def primaryDiagnoses = wrapper.getUniqueDiagnoses(true, false)
+            def primaryDiagnoses = wrapper.getUniqueDiagnoses(true, false);
+           
+            for (def encounter : wrapper.getSortedEncounters()){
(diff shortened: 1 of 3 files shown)


omod/src/main/webapp/fragments/patientdashboard/visits.gsp
```diff
@@ -70,8 +70,15 @@
...
         <%
             def visits = patient.allVisitsUsingWrappers;
+
             visits.eachWithIndex { wrapper, idx ->
-                def primaryDiagnoses = wrapper.getUniqueDiagnoses(true, false)
+            def primaryDiagnoses = wrapper.getUniqueDiagnoses(true, false);
+           
+            for (def encounter : wrapper.getSortedEncounters()){
+               def  newdiagnoses = diagnosisService.getPrimaryDiagnoses(encounter);
+               primaryDiagnoses << [newdiagnoses]
```
Ruhanga: @reagan-meant, I guess this works best with ordinary arrays than with lists. I would suppose you do something like `primaryDiagnoses.addAll(newdiagnoses)`.

reagan-meant: You are right...I came to the same conclusion after checking up on the different flavours of <<, addAll, "+" and push that groovy offers.....Will be making an updated PR

reagan-meant:
Been able to solve it....The issue was with how I was adding new ArrayLists to the existing one.

PR #309, opened 2020-04-23
RA-1737: <programstatus/> to maximise/minimise states based on param.

Ruhanga:
Ticket: https://issues.openmrs.org/browse/RA-1737

### DIFF
--- a/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
+++ b/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
@@ -29,7 +29,7 @@ export default class ProgramStatusController {
         this.programOutcomes = null;
         this.sessionLocation = null;
 
-        this.expandedWorkflow = typeof this.config.expandedWorkflow === 'boolean' ?  this.config.expandedWorkflow : false;
+        this.expandedWorkflows = typeof this.config.expandedWorkflows === 'boolean' ?  this.config.expandedWorkflows : false;
 
         // we calculate these two values based on the completion date of the previous program and the enrollment date of any subsequent program
         // we populate these on initial load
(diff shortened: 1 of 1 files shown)


omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
```diff
@@ -148,13 +150,14 @@ export default class ProgramStatusController {
...
-            this.expanded.workflow[workflowUuid] = true;
-        }
-        else {
-            this.expanded.workflow[workflowUuid] = !this.expanded.workflow[workflowUuid];
-        }
+        this.expanded.workflow[workflowUuid] = !this.expanded.workflow[workflowUuid];
+    }
+
+    // Sets workflowws to expanded display of containing states based on boolean configuration value - this.config.expandedWorkflow
+    expandWorkflows() {
```
mogoodrich, line 157: This method is really "setInitialExpandedStateForAllWorkflows", correct?  That's bit wordy, happy if we can come up with something shorter, but "expandWorkflows" seems incorrect and confusing (if I'm understanding what this is doing correctly, let me know if I'm wrong!)

mks-d: @Ruhanga there's also a typo in the comment above the new method ("workflowws").

Btw could this not have worked:
```javascript
toggleExpandedWorkflow(workflowUuid) {
    // the first time we hit this, we need to initialize the workflow
    if (!workflowUuid in this.expanded.workflow) {
        this.expanded.workflow[workflowUuid] = this.expandedWorkflow;
    }
    else {
        this.expanded.workflow[workflowUuid] = !this.expanded.workflow[workflowUuid];
    }
}
```
?

Also I would probably rename `this.expandedWorkflow` into something like `this.initExpandedWorkflow`.

Ruhanga: @mks-d, the above snippet executes when the `Show more` is first clicked and later. So upon first page loading this doesn't get hit.

Thanks, I'll rename `this.expandWorkflows` to `this.initExpandedWorkflow`. I guess that would be consistent with `this.toggleExpandedWorkflow`. cc: @mogoodrich.

mogoodrich: Yes, I like "initExpandedWorkflows", thanks @Ruhanga

PR #333, opened 2020-07-21
Remove concept list size limit

brandones:
It's an arbitrary and undocumented limitation on implementers. Instead of having this limitation, a note should be added to the [wiki documentation](https://wiki.openmrs.org/display/docs/Patient+Summary+Widget+Documentation#PatientSummaryWidgetDocumentation-conceptlistLatestObsForConceptList) which advises implementers that adding too many concepts can impact performance.

### DIFF
--- a/omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
+++ b/omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
@@ -6,33 +6,19 @@ export default class LatestObsForConceptListController {
     }
 
     $onInit() {
-        this.maxConceptCount = 10;
         this.maxAgeInDays = undefined;
         this.obs = [];
-
         this.openmrsRest.setBaseAppPath("/coreapps");
-
         this.maxAgeInDays = this.widgetsCommons.maxAgeToDays(this.config.maxAge);
 
-        // Remove whitespaces
-        let concept_list = this.config.concepts.replace(/\s/g,'');
         // Number of latest observations
         let n_LatestObs = !angular.isUndefined(this.config.nLatestObs) && this.config.nLatestObs != null ? this.config.nLatestObs : 1;
-        // remove all concepts over the maximum number of concepts
-        let concepts = this.config.concepts.split(",");
-        // from the number of elements specified by maxConceptCount remove the remaining elements
-        let elements_to_remove = concepts.length - this.maxConceptCount;
-        if (elements_to_remove > 0) {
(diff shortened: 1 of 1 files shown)


omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
```diff
@@ -6,33 +6,19 @@ export default class LatestObsForConceptListController {
...
-        this.maxConceptCount = 10;
         this.maxAgeInDays = undefined;
         this.obs = [];
-
         this.openmrsRest.setBaseAppPath("/coreapps");
-
         this.maxAgeInDays = this.widgetsCommons.maxAgeToDays(this.config.maxAge);
 
-        // Remove whitespaces
-        let concept_list = this.config.concepts.replace(/\s/g,'');
```
mogoodrich, line 18: Can we confirm we don't need this?  ie confirm that the latestobs widget properly strips out whitespace?

brandones: The code as written actually doesn't strip whitespace. You can see that the `concept_list` variable just gets replaced by something else entirely on line 29. No one seems to have noticed the change in behavior, and I can't actually think of a reason why you would want to strip whitespace from the items in the concept list. Maybe the intention is to `trim`, so that it tolerates whitespace around the comma? Yeah actually I'll add that in; that would make sense to me.

mogoodrich: yes I was assuming it was to make sure if you had a list like "123, abc, def" it didn't parse it as "123", " abc", " def" and then not find concepts " abc" and " def"

omod/src/main/web/dashboardwidgets/latestobsforconceptlist/latestobsforconceptlist.controller.js
```diff
@@ -6,33 +6,19 @@ export default class LatestObsForConceptListController {
...
-        }
-
-        concept_list = concepts.join(",");
 
         // Fetch last obs or obsGroup for the list of concepts
         this.openmrsRest.list('latestobs', {
             patient: this.config.patientUuid,
             v: 'full',
-            concept: concept_list,
+            concept: this.config.concepts.split(',').map(c => c.trim()).join(','),
```
mogoodrich, line 21: Cool, yeah, this is what I was thinking it was there for...

PR #336, opened 2020-07-22
For Latest Obs For Concept List, make concept name type configurable

brandones:
This includes some refactoring and code cleanup. I'll point out the functional changes.

Screenshot from 2020-07-22 14-09-14
Screenshot from 2020-07-22 14-08-47

PR #337, opened 2020-07-25
Refactor obsacrossencounters controller; add useConceptNameForValue config option

brandones:
PIH ticket https://pihemr.atlassian.net/browse/UHM-4499

This includes very little functional change. I will indicate it in the comments.

Definitely use "hide whitespace changes" when reviewing.

It should be noted that although the [Documentation](https://wiki.openmrs.org/display/docs/Patient+Summary+Widget+Documentation#PatientSummaryWidgetDocumentation-conceptencountersObsAcrossEncounters) claims that `concepts` can be specified by ID, UUID, FSN, or Map, it's clear from the code that only UUID could possibly have worked. I haven't attempted to fix that in this PR.

### DIFF
--- a/omod/src/main/web/dashboardwidgets/dashboardwidgetscommons.service.js
+++ b/omod/src/main/web/dashboardwidgets/dashboardwidgetscommons.service.js
@@ -123,7 +123,7 @@ export default class WidgetsCommons {
     }
 
     isDrug(obj) {
-        return !!obj.concept
+        return Boolean(obj.concept);
     }
 
 }
--- a/omod/src/main/web/dashboardwidgets/dashboardwidgetscommons.service.js
+++ b/omod/src/main/web/dashboardwidgets/dashboardwidgetscommons.service.js
@@ -93,6 +93,7 @@ export default class WidgetsCommons {
      * 
      * @param {object} concept Must have rep including `display,names:(voided,locale,conceptNameType,localePreferred,name)`
      * @param {string} nameType One of "FSN", "shortName", "preferred"
+     * @param {string} locale The preferred locale
      */
     getConceptName(concept, nameType, locale) {
         const names = concept.names.filter(n => !n.voided && n.locale === locale);
@@ -110,4 +111,19 @@ export default class WidgetsCommons {
         }
         return resultName ? resultName.name : concept.display;
     }
+
+    hasDatatypeDateOrSimilar(concept) {
+        return ['8d4a505e-c2cc-11de-8d13-0010c6dffd0f',
+            '8d4a591e-c2cc-11de-8d13-0010c6dffd0f',
+            '8d4a5af4-c2cc-11de-8d13-0010c6dffd0f'].includes(concept.datatype.uuid)
+    }
+
+    hasDatatypeCoded(concept) {
+        return concept.datatype.uuid === "8d4a48b6-c2cc-11de-8d13-0010c6dffd0f";
+    }
+
+    isDrug(obj) {
+        return !!obj.concept
+    }
+
 }
--- a/omod/src/main/web/dashboardwidgets/dashboardwidgetscommons.service.js
+++ b/omod/src/main/web/dashboardwidgets/dashboardwidgetscommons.service.js
@@ -1,90 +1,113 @@
 export default class WidgetsCommons {
 
-        // Method used to define days since given date and return it as message
-        dateToDaysAgoMessage(date) {
-            let days = this.dateToDaysAgo(date);
+    // Method used to define days since given date and return it as message
+    dateToDaysAgoMessage(date) {
+        let days = this.dateToDaysAgo(date);
 
-            if (days == 0) {
-                return "today"
-            }
-            else if (days == 1) {
-                return days + " day ago"
-            }
-            else {
-                return days + " days ago"
-            }
-        };
+        if (days == 0) {
+            return "today"
+        }
+        else if (days == 1) {
+            return days + " day ago"
+        }
+        else {
+            return days + " days ago"
+        }
+    };
 
-        //Number of days since the given date
-        daysSinceDate(date) {
+    //Number of days since the given date
+    daysSinceDate(date) {
+        let days = 0;
+        if (date !== null) {
+            let givenDate = new Date(date).setHours(0,0,0,0);
+            let today = new Date().setHours(0,0,0,0);
+            let seconds = Math.floor((today - givenDate) / 1000);
+            days = Math.floor(seconds / 86400);
+        }
+
+        return days;
+    };
+
+    // Method used to define days since given date
+    dateToDaysAgo(date) {
+        const time = new Date(date).getTime();
+        const seconds = Math.floor((new Date().getTime() - time) / 1000);
+        const interval = Math.floor(seconds / 86400);
+        let days = 0;
+        if (interval > 1) {
+            days = interval;
+        }
+        return days;
+    };
+
+    // Method used to parse maxAge to day count
+    maxAgeToDays(maxAge) {
+        if (typeof maxAge === 'undefined') {
+            return undefined;
+        } else {
             let days = 0;
-            if (date !== null) {
-              let givenDate = new Date(date).setHours(0,0,0,0);
-              let today = new Date().setHours(0,0,0,0);
-              let seconds = Math.floor((today - givenDate) / 1000);
-              days = Math.floor(seconds / 86400);
-            }
-
-            return days;
-        };
-
-        // Method used to define days since given date
-        dateToDaysAgo(date) {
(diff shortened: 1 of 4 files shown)


omod/src/main/web/dashboardwidgets/dashboardwidgetscommons.service.js
```diff
@@ -1,90 +1,113 @@
...
-            }
-            return today;
-        };
+    /**
+     * Returns the most appropriate name given a concept and a name type preference
+     * 
+     * @param {object} concept Must have rep including `display,names:(voided,locale,conceptNameType,localePreferred,name)`
+     * @param {string} nameType One of "FSN", "shortName", "preferred"
+     */
+    getConceptName(concept, nameType, locale) {
```
brandones, line 97: brought over from latestobsforconceptlist

omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
```diff
@@ -29,177 +28,117 @@ export default class ObsAcrossEncountersController {
...
+          this.simpleEncs.push({
+            encounterType: encounter.encounterType.name,
+            encounterDatetime: encounter.encounterDatetime,
+            obs: obsRow
+          });
+        }
       }
       // if the widget was configured to display the obs concept SHORT name instead of the default obs.value.display value
       // By default, the widget displays the obs.value.display property
-      if (this.config.useConceptShortName && this.config.useConceptShortName === 'true') {
```
brandones, line 199: This is a functional change. Previously `useConceptShortName` only worked if provided the string value `"true"`. It now works for `true`. It will still work for `"true"`, so it isn't a breaking change (unless anyone specified the string `"false"`, but that seems unlikely).

omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
```diff
@@ -29,177 +28,117 @@ export default class ObsAcrossEncountersController {
...
-      }
+  getObsValue(obs) {
+    if (['8d4a505e-c2cc-11de-8d13-0010c6dffd0f',
+        '8d4a591e-c2cc-11de-8d13-0010c6dffd0f',
+        '8d4a5af4-c2cc-11de-8d13-0010c6dffd0f'].includes(obs.concept.datatype.uuid)) {
+        // If value is date, time or datetime
+        return this.$filter('date')(new Date(obs.value), this.config.dateFormat);
+    } else if (angular.isDefined(obs.value.display)) {
+        // If value is a concept
+        return this.config.useConceptNameForValue ? obs.value.concept.display : obs.value.display
```
brandones, line 90: This adds support for the config value `useConceptNameForValue`.

omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
```diff
@@ -29,177 +28,117 @@ export default class ObsAcrossEncountersController {
...
-              let enc = {
-                encounterType: encounter.encounterType.name,
-                encounterDatetime: encounter.encounterDatetime,
-                obs: foundObs
-              };
-              this.simpleEncs.push(enc);
-            } else {
-              // we have an incomplete match, just add the values to the existing object
-              for (key in Object.keys(foundObs)){
-                searchObs[key] = searchObs[key];
```
brandones, line 182: I kid you not

brandones: At least this code was logically unreachable

omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
```diff
@@ -29,177 +28,107 @@ export default class ObsAcrossEncountersController {
...
-              }
```
brandones, line 183: I don't *think* there's a functional change here. I'm pretty sure this code was unreachable, and it definitely does nothing. But it's hard to say how the function used to work, altogether.

omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
```diff
@@ -209,44 +138,20 @@ export default class ObsAcrossEncountersController {
...
-              obs.value.display = shortDisplay.display;
-            }, function (err) {
-              console.log(`failed to retrieve concept ${conceptUuid}, ${err}`);
-            });
-          }
+    for (let encounter of encounters) {
+      for (let obs of encounter.obs) {
+        // if the obs value is coded then we try to look for the short names of the coded answers
+        // some obs.value nodes contain the concept property that points to the concept
+        // and other coded obs(e.g. Drug Frequency) do not contain a child concept property
```
mogoodrich, line 145: The distinction here is whether the obs is valueCoded (so the value is a concept) or valueDrug (so the value is a drug), so whether obs.value = Concept or obs.value = Drug.

Drug objects have a concept property, so therefore if obs.value = Drug, then obs.value.concept.uuid is the concept associated with that drug.  So what I believe the chunk of code below is doing is saying "if this is a valueCoded, set the obs.value.display to the shortname of the concept, and if this is a valueDrug, set the obs.value.display to the shortname of the concept associated with the drug (when by default I believe it is the name of the drug)"

omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
mogoodrich: Read by comment below first, but I don't think this is what we want here... the idea is, "if this obs is a valueDrug, do we want to use the name of the drug, or the name of the concept associated with the drug"?  I believe "obs.value.concept.display" will result in a NPE unless this is a "valueDrug" obs.

brandones: Oops, good catch about the NPE. But isn't `obs.value.concept.display` also present for value coded?

omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
brandones: Yyyyyes. This seems like what we want. Right?

mogoodrich: Yeah, I just thought the comment was a confusing and since what is happening here is confusing I wanted to make srue the comment was clear:  the difference is between obs that are "valueCoded" (the value is an Obs) and "valueDrug" (the value is a Drug).    In the 95% use care for the widget, will likely just be displaying "valueCoded"... this code is just to handle the 5% use case (which happens to be our use case here) where the value is is a valueDrug.  The comment makes it sound arbitrary ("some obs.value nodes contain the concept... and other coded obs ... do not").  I think there was confusion when this code was originally written why it needed to be that way, and I wanted to clear that up.

omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
mogoodrich: I don't think so... in valueCoded, "obs.value" = Concept, and a "concept" doesn't have a concept property.

Take a look at the screenshot on the ticket, and note that the object that obs.value points to is a drug, with all the drug fields like "maxDailyDose", etc....

mks-d:
Same here, @brandones @mogoodrich we are still assessing what's going exactly but this PR seems to be breaking stuff in our environment.

We have broken 'obsacrossencounters' and 'latestobsforconceptlist' widgets all of a sudden following the current PR and [this one](https://github.com/openmrs/openmrs-module-coreapps/pull/336).

sherrif10:
We have broken 'obsacrossencounters' and 'latestobsforconceptlist' widgets all of a sudden following the current PR. i agree with @mks-d, obsAcrossEncounter widget seems to have been broken, however am reviewing this code to see where it was broken and fix RA-1800

brandones:
@sherrif10 I think someone from @mks-d 's team is already working on it; you may want to sync up with them.

sherrif10:
Thanks @brandones , am working on that ticket still, since you refactored alot of code, it needs me to go it deeply to understand the code format to resolve,so am inquiring some time to see way forward

brandones:
@sherrif10 , could you describe the breakage (screenshots would be extremely helpful), and if possible provide the app configuration for which it breaks? I can probably help identify what's going wrong.

PR #338, opened 2020-07-27
RA-1795: <encounterDiagnoses/> to skip voided diagnoses.

icrc-psousa:
Related ticket: https://issues.openmrs.org/browse/RA-1795

### DIFF
--- a/api-2.2/src/main/java/org/openmrs/module/coreapps/htmlformentry/EncounterDiagnosesElement.java
+++ b/api-2.2/src/main/java/org/openmrs/module/coreapps/htmlformentry/EncounterDiagnosesElement.java
@@ -24,7 +24,6 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.StringTokenizer;
-import java.util.stream.Collectors;
 
 import javax.servlet.http.HttpServletRequest;
 
@@ -381,10 +380,17 @@ public class EncounterDiagnosesElement implements HtmlGeneratorElement, FormSubm
         if (mode == FormEntryContext.Mode.EDIT || mode == FormEntryContext.Mode.VIEW) {
             Encounter existingEncounter = context.getExistingEncounter();
             if (existingEncounter != null) {
-                return existingEncounter.getDiagnoses().stream().filter(d -> !d.getVoided()).collect(Collectors.toSet());
+                Set<org.openmrs.Diagnosis> nonVoidedDiagnoses = new HashSet<org.openmrs.Diagnosis>();
+                for (org.openmrs.Diagnosis diagnosis : existingEncounter.getDiagnoses()) {
+                    if (!diagnosis.getVoided()) {
+                        nonVoidedDiagnoses.add(diagnosis);
+                    }
+                }
+                return nonVoidedDiagnoses;
             }
         }
-        return new HashSet<>();
+
+        return new HashSet<org.openmrs.Diagnosis>();
     }
 
     private List<Diagnosis> getPriorDiagnoses(FormEntryContext context, DispositionType dispositionType) {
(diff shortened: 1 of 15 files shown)


api-2.2/src/main/java/org/openmrs/module/coreapps/htmlformentry/EncounterDiagnosesElement.java
```diff
@@ -378,10 +378,16 @@ public void applyAction(FormEntrySession formEntrySession) {
...
-        	}
+            Encounter existingEncounter = context.getExistingEncounter();
+            if (existingEncounter != null) {
+                Set<org.openmrs.Diagnosis> nonVoidedDiagnoses = new HashSet<org.openmrs.Diagnosis>();
+                for (org.openmrs.Diagnosis diagnosis : existingEncounter.getDiagnoses()) {
+                    if (!diagnosis.getVoided()) {
+                        nonVoidedDiagnoses.add(diagnosis);
+                    }
+                }
+                return nonVoidedDiagnoses;
```
mks-d, line 389: I was hoping we could do this:
```java
return existingEncounter.getDiagnoses().stream().filter(d -> !d.getVoided()).collect(Collectors.toSet());
```
But for this to work we would need to put the JDK compliance to 1.8 (_just_ for the api-2.2 submodule). Do you think you could try that out?

icrc-psousa: I tried just to specify JDK 1.8 in the api-2.2 module but, all existing tests relying Spring ApplicationContext began to fail. It seems to be using Spring 3 thus failing to load Java 8 classes.

```
java.lang.IllegalStateException: Failed to load ApplicationContext
(...)
Caused by: org.springframework.core.NestedIOException: ASM ClassReader failed to parse class file - probably due to a new Java class file version that isn't supported yet
```

icrc-psousa: @mks-d, api-2.2 submodule is now using JDK 8 and I've reviewed the code to use streams.

icrc-psousa:
@mks-d, actually I had updated the pull request before testing openmrs at runtime.
When starting openmrs with the updated module, ApplicationContext seems to be failing to initializing. I'll still take a more careful look.

`
openmrs-web_1    | Related cause: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'appFrameworkService' defined in URL [jar:file:/opt/openmrs/.openmrs-lib-cache/appframework/lib/appframework-api-2.16.0-SNAPSHOT.jar!/moduleApplicationContext.xml]: Cannot create inner bean 'org.openmrs.module.appframework.service.AppFrameworkServiceImpl#350f5214' of type [org.openmrs.module.appframework.service.AppFrameworkServiceImpl] while setting bean property 'target'; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'org.openmrs.module.appframework.service.AppFrameworkServiceImpl#350f5214' defined in URL [jar:file:/opt/openmrs/.openmrs-lib-cache/appframework/lib/appframework-api-2.16.0-SNAPSHOT.jar!/moduleApplicationContext.xml]: Cannot resolve reference to bean 'allComponentsState' while setting constructor argument; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'allComponentsState' defined in URL [jar:file:/opt/openmrs/.openmrs-lib-cache/appframework/lib/appframework-api-2.16.0-SNAPSHOT.jar!/moduleApplicationContext.xml]: Initialization of bean failed; nested exception is org.springframework.aop.framework.AopConfigException: Unexpected AOP exception; nested exception is java.lang.ClassCastException: org.openmrs.module.appframework.repository.AllComponentsState$$EnhancerBySpringCGLIB$$9916589d cannot be cast to org.springframework.cglib.proxy.Factory
`

PR #339, opened 2020-07-29
Fix obs datetime

brandones:
Fixes an error introduced in https://github.com/openmrs/openmrs-module-coreapps/pull/336/files . I changed from `full` obs rep to a custom rep, but neglected to include `obsDatetime` in that rep.


brandones:
trivial; merging

PR #344, opened 2020-09-07
RA-1822 Local variables should not be declared and then immediately returned

achilep:
issue I work on : https://issues.openmrs.org/browse/RA-1822


sacull (approved):
It looks good to me. ;)
@mogoodrich @dkayiwa

PR #345, opened 2020-09-07
RA-1823 Assertion arguments should be passed in the correct order

achilep:
Issue I work on : https://issues.openmrs.org/browse/RA-1823


sacull (approved):
It looks good to me. ;)
@mogoodrich @dkayiwa

PR #350, opened 2020-09-10
RA-1834:Comparison of String objects using == or !=

Wandji69:
Using the ```equal(Object)``` to test equality between objects and strings.


Wandji69:
https://issues.openmrs.org/browse/RA-1834

sacull (approved):
It looks good to me. ;)
@mogoodrich @dkayiwa

PR #352, opened 2020-09-12
RA-1828: Return of boolean expressions should not be wrapped into an "if-then-else" statement

DBryzz:
[RA-1828](https://issues.openmrs.org/browse/RA-1828): Return of boolean expressions should not be wrapped into an "if-then-else" statement

Replaced if-then-else statement with a boolean expression. [Issue description](https://issues.openmrs.org/browse/RA-1828).

Please review and merge if all is ok.

### DIFF
--- a/api/src/main/java/org/openmrs/module/coreapps/parser/ParseEncounterToJson.java
+++ b/api/src/main/java/org/openmrs/module/coreapps/parser/ParseEncounterToJson.java
@@ -93,11 +93,7 @@ public class ParseEncounterToJson {
             simpleEncounter.put("canDelete", true);
         }
 
-        if (verifyIfUserHasPermissionToEditAnEncounter(encounter, authenticatedUser, canEdit)) {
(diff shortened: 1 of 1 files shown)


sacull:
@DBryzz Why not `simpleEncounter.put("canEdit", verifyIfUserHasPermissionToEditAnEncounter(encounter, authenticatedUser, canEdit));`?

DBryzz:
Just thought it's more presentable this way
but if you say I should change it then I will

sacull:
In my opinion, creating a variable that is named almost the same as a method whose result is assigned to that variable is an unnecessary duplication of elements in the code.

DBryzz:
Alright, I'll change it.
Thanks for the correction

DBryzz:
@sacull changes made

sacull (approved):
In my opinion, it is good now. ;)

Wandji69 (approved):
This looks good now. Thanks @DBryzz.

PR #356, opened 2020-09-15
RA-1800 Obs Across Encounter widget broken

brandones:
https://issues.openmrs.org/browse/RA-1800

Fixes a typo which caused ObsAcrossEncounter not to work

@sherrif10 I was unable to test that this actually fixes the problem, because I can't make any sense of "Manage Apps" and the widget "config"s. Please try testing using this branch and let me know if it fixes the problem. Sorry for the inconvenience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,13 @@ Provides apps for performing the common tasks
 
 # Development
 
-As usual, you can just use `mvn openmrs-sdk:watch` for development.
+Hot reloading changes is supported. Use the following three commands:
 
-However, in order to see changes to `scss` files, you'll also need to run
-`mvn process-resources -Pwatch-sass`.
+```
+mvn openmrs-sdk:watch
+mvn process-resources -Pwatch-sass
+npm run watch
+```
 
 ## Dashboard Widgets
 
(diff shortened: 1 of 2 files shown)


sherrif10:
Thanks actually i had a same working work like this in my IDE  though it kept failing thats why i delayed pushing a Pr when i tested, let me test this and see, i will report to you back

sherrif10 (commented):
Let me test it out however i had similar finding, but i will get back to you today

mogoodrich (approved):

mogoodrich:
Thanks @brandones .... also @mks-d  I don't know if you have an easy way to test with this PR?  It looks like a clear error, so probably makes sense to just go ahead and merge?

sherrif10:
@brandones  @mogoodrich  still this did not resolve the error i tried to work it like [ https://github.com/sherrif10/openmrs-module-coreapps/commit/0496a297679a84815e05050f6661a36803b930c7](url)

brandones:
It's working for me.

sherrif10:
Thanks @brandones , it works fine with me now, we are ready to go with this

PR #360, opened 2020-09-21
RA-1530 Program List dashboard links don't behave like real links: take 2

brandones:
This fixes an issue introduced in 4546d08e7ac2bf85978b947c84822efc0e8f8cba
which causes the "Enroll" button not to work.

https://issues.openmrs.org/browse/RA-1530


mogoodrich (approved):
Oops... yeah, fix LGTM.  I assume you searched for other uses of gotoProgramDashboard at this point?

brandones:
Yep, there's only those two. Thanks Mark.

brandones:
Assuming I only have to look in coreapps and not elsewhere.

PR #361, opened 2020-09-21
RA-1840 latestobsforconceptlist widget broken

brandones:
https://issues.openmrs.org/browse/RA-1840


mogoodrich (approved):
Haven't studied in detail, but generall LGTM... I assume the issue that Ines was reporting was because of the missing "value" for group members, which I guess would be for group members that have numeric values?

brandones:
`value` was missing for any kind of value. But the first thing that would happen is that something would hit `obs.value.display` for the obsgroup, which would crash the script because `value` is undefined.

PR #374, opened 2020-11-26
RA-1874: Added variable tracking that Enroll button is clicked.

icrc-toliveira:
Ticket ID:
https://issues.openmrs.org/browse/RA-1874

Problem:
Its possible to press the enroll button multiple times and as consequence, enroll the patient multiple times

Description:
Added variable requestInProgress to disable the Enroll Button while the request is being processed

### DIFF
--- a/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
+++ b/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
@@ -331,6 +331,7 @@ export default class ProgramStatusController {
             }).then((response) => {
                 if (this.config.disableReloadPage) {
                     this.fetchPatientProgram(this.patientProgram.uuid); // refresh display
+                    this.requestInProgress = false;
                 }
                 else {
                     this.reloadPage();  // closing a program affects other widgets, so we need to reload the entire page
--- a/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
+++ b/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
@@ -47,6 +47,7 @@ export default class ProgramStatusController {
         this.markPatientDeadOutcome = null;
         this.markPatientDeadPage = "/coreapps/markPatientDead.page?patientId={{patientUuid}}&returnDashboard={{dashboard}}&defaultDeathDate={{date}}&defaultDead=true";
         this.canMarkPatientDead = false;
+        this.requestInProgress = false;
 
         // config parameter that can be passed in so the widget knows what dashboard it is being rendered on,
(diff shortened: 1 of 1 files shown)


omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
```diff
@@ -313,6 +314,7 @@ export default class ProgramStatusController {
 
     enrollInProgram() {
         if (this.input.dateEnrolled && this.input.enrollmentLocation) {
+            this.requestInProgress = true;
```
mks-d, line 317: @icrc-toliveira when is this ever reset to false when the request has finished processing? It looks like it's only happening once initially in `$onInit()` as of now.

mogoodrich (commented):
Looks like @mks-d 's comment needs to be addressed, but I'm in favor of this functionality once/if it works properly!

omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
jwnasambu: @icrc-ifernandes kindly your ticket id is `RA-1874:Enrolment button on "Program Status"  can be pressed multiple times`  and not` RA-1874: Added variable to prevent multiple click on Enroll Button`. Kindly fix it.

mks-d: @jwnasambu the PR headline does not have to match the ticket's headline.
* The ticket headline describes a certain scope of work.
* The PR headline describes a change made to the code.

The issue can be described as "the enroll button can be clicked twice and it shouldn't" and the code change can be described as "adding a variable to greyline the enroll button".

omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
```diff
@@ -329,6 +331,7 @@ export default class ProgramStatusController {
             }).then((response) => {
                 if (this.config.disableReloadPage) {
                     this.fetchPatientProgram(this.patientProgram.uuid); // refresh display
+                    this.requestInProgress = false;
                 }
                 else {
                     this.reloadPage();  // closing a program affects other widgets, so we need to reload the entire page
```
mks-d, line 337: @mogoodrich actually I missed that but @icrc-toliveira pointed out that a subsequent reset of `this.requestInProgress` to `false` is made within `this.reloadPage()`.

So looks like it's ok to me 👍

mks-d (approved):

mogoodrich (approved):
LGTM, thanks!

PR #384, opened 2021-03-01
RA-1747:Remove or hide admit button

HerbertYiga:
ticket Id:https://issues.openmrs.org/browse/RA-1747

Description
Remove or hide admit button

still in progress


omod/src/main/webapp/pages/clinicianfacing/patient.gsp
```diff
@@ -118,6 +118,7 @@ ${ ui.includeFragment("coreapps", "patientHeader", [ patient: patient.patient, a
                     <h3 >${ ui.message("coreapps.clinicianfacing.activeVisitActions") }</h3>
                     <% visitActions.each { ext -> %>
                     <li class="float-left">
+                    <% if(ext.label) != "Admit to inpatient" %>
```
dkayiwa, line 121: Will this work if the label is in a language other than English?

HerbertYiga: @dkayiwa  let me re -think the solution, also on the ticket description i see some thing like (Enters ADT Events role should theoretically not show the "Admit to inpatient" option for Standard Users) does this mean we disable the admin to inpatient for standard users? If true,i am requesting for some insights about the standard users!!

mogoodrich: Also, we shouldn't be removing anything here, at the UI level... this seems more like some sort of ref app configuration issue?

HerbertYiga: thanks @mogoodrich  looking at this ticket Id:https://issues.openmrs.org/browse/RA-1747,the ticket description is some how  unclear to me on (when) the admit to patient button is not needed, this can give me an easy move on the (where) to make the fix,

PR #387, opened 2021-03-09
RA-1864: Handling undefined encounterType and encounterTypes attribut…

NischithJShetty:
…es in the obsacrossencounters widget configuration

This Pull Request fixes the JIRA ticket: https://issues.openmrs.org/browse/RA-1864

The attributes  “encounterType” and “encounterTypes” are now optional while configuring the widget “obsacrossencounters” (and they  need not be defined in the widget config file: “/coreapps-omod/src/main/resources/apps/dashboardWidgets_app.json”), the file “/coreapps-omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js” has been altered to handle the logic for ‘null’ values.

### DIFF
--- a/omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
+++ b/omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
@@ -55,7 +55,7 @@ export default class ObsAcrossEncountersController {
 
   fetchEncounters() {
     const encounterTypes = this.config.encounterTypes ? this.config.encounterTypes.split(',').map(c => c.trim()) : [];
-    const legacyEncounterTypes = this.config.encounterType ? this.config.encounterType.split(',').map(c => c.trim()) : [];
+    const legacyEncounterTypes = this.config.encounterType ? this.config.encounterType.split(',').map(c => c.trim()) : ["123AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"];
     encounterTypes.push(...legacyEncounterTypes);
 
     const encounterPromises = encounterTypes.map(e =>
--- a/omod/src/main/resources/apps/dashboardWidgets_app.json
+++ b/omod/src/main/resources/apps/dashboardWidgets_app.json
@@ -33,7 +33,6 @@
       "icon": "icon-user-md",
       "label": "coreapps.clinicianfacing.healthTrendSummary",
       "concepts": "5087AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA,5088AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA,5089AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
-      "encounterType": "123AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
       "maxRecords": "5",
       "maxAge": "1m"
     },


omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
```diff
@@ -55,7 +55,7 @@ export default class ObsAcrossEncountersController {
 
   fetchEncounters() {
     const encounterTypes = this.config.encounterTypes ? this.config.encounterTypes.split(',').map(c => c.trim()) : [];
-    const legacyEncounterTypes = this.config.encounterType ? this.config.encounterType.split(',').map(c => c.trim()) : [];
+    const legacyEncounterTypes = this.config.encounterType ? this.config.encounterType.split(',').map(c => c.trim()) : ["123AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"];
```
dkayiwa, line 58: Does this mean that when an encounter type is not provided, we use a hard coded one?

brandones:
Sorry I missed this. This broke a bunch of tests, as well as breaking the contract for the widget. If no encounter type is provided in the configuration, the widget should display results from *any encounter type*, not only results from one arbitrary encounter type that happens to work for a particular implementation.

Both of those tickets appear to have been filed in between when I completely broke this widget in Fall 2020 and when I wrote a bunch of tests and fixed it on March 12 (f03edbdca427c0e44e40e5318db7db357c04cc0c).

I'm inclined to revert this. And then Stephen and Grace should check whether the issues they found are still issues. And then if they are, we should represent those issues as *additional* tests, and then fix those in a way that doesn't break everything else.

Sound good, @dkayiwa ?

dkayiwa:
@brandones sounds good to me.

NischithJShetty:
@brandones There was very limited documentation for me to work on this ticket and so there were some limitations. The only documentation that was available was [this](https://wiki.openmrs.org/display/docs/Patient+Summary+Widget+Documentation)

I see that now this commit has been reverted and the below line has been added to the "obsacrossencounters" widget configuration:
"encounterType": "123AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"

Could you please elaborate on what the value "123AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" stands for here and what are the other possible values that the attribute "encounterType" can be assigned with. Sorry to bother you as I could not find any documentation in this regard.

Thank you.

brandones:
Hi @NischithJShetty . `123AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA` is evidently the UUID of an encounter type. It is one that comes bundled with the RefApp, which is why its used in the definition of the default ObsAcrossEncounters widget. Please see [the documentation](https://wiki.openmrs.org/display/docs/Patient+Summary+Widget+Documentation#PatientSummaryWidgetDocumentation-conceptencountersObsAcrossEncounters) for information about usage of this widget, including its configuration parameters.

NischithJShetty:
Thanks @brandones

The documentation you have mentioned is the one I had referred to as well. The confusing bit there was, it does not detail the possible values (& the description of the values) that can be assigned the "encounterType" parameter.

brandones:
Sure, I've updated the docs to say more explicitly that an encounter type UUID is what is expected. Though it is part of the example JSON definition as well. Thanks for your feedback.

NischithJShetty:
Hi @brandones 

Please help me with one more clarification. I tried to understand the possible UUID values that can be assigned to "encounterType" during the ObsAcrossEncounters widget configuration. So, checked the values of UUIDs from the "encounter_type" table of the openmrs database and got the below output:

image

"123AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" does not seem to be one of the UUID values in the table.

Not sure what I am missing here. Thank you.

brandones:
That's true. That's the RefApp metadata you have there, and it doesn't have that UUID. That UUID actually looks like a CIEL concept UUID, which certainly wouldn't make sense as an encounter type UUID.

I doubt that that default configuration is working for anyone. Fortunately there are two tickets, https://issues.openmrs.org/browse/RA-1864 and https://issues.openmrs.org/browse/RA-1871 , which have outstanding requests for testing from @ssmusoke and @gracepotma .

PR #385, opened 2021-03-05
RA-1899 pt 1: Write tests for latestobsforconceptlist

brandones:
https://issues.openmrs.org/browse/RA-1899

This also improves some UI text in the latestobsforconceptlist widget.

Screenshot from 2021-03-04 16-03-04

PR #388, opened 2021-03-10
RA-1899 Part 2 - Tests for obsacrossencounter patient summary widget

brandones:
https://issues.openmrs.org/browse/RA-1899

Part one was here: https://github.com/openmrs/openmrs-module-coreapps/pull/385/

Hopefully these tests can serve as a template for tests for the other dashboardwidgets. I know visitbyencountertype is under active development, CC @mks-d .

The key to testing these is to make the controller concentrate all of its output in an `output` object, and make the HTML file as simple as possible (by concentrating logic in the controller). A bit of refactoring may be necessary to make the `$onInit` function of the controller return a promise that only resolves once the correct `output` has been produced. Tests then touch only three points: the widget config, the REST calls, and the output.

This also fixes a couple bugs with ObsAcrossEncounters.

### DIFF
--- a/omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
+++ b/omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
@@ -6,21 +6,24 @@ export default class ObsAcrossEncountersController {
   }
 
   $onInit() {
-    this.order = 'desc';
     this.sessionLocale = null;
     this.CONCEPT_CUSTOM_REP = 'custom:(uuid,display,names:(voided,locale,conceptNameType,localePreferred,name)';
-
-    // a map of conceptUUID --> concept(REST response)
-    this.conceptsMap = {};
+    this.conceptsMap = {};  // a map of conceptUUID --> concept(REST response)
     this.simpleEncs = [];
     this.headers = [];
     this.openmrsRest.setBaseAppPath("/coreapps");
 
+    this.output = {
+      headers: [],
+      rows: [],
+      maxAgeInDays: this.widgetsCommons.maxAgeToDays(this.config.maxAge)
+    }
+
     this.fetchSessionInfo();
-    this.fetchHeaders();
-    this.fetchConcepts();
-    this.fetchEncounters();
-    this.maxAgeInDays = this.widgetsCommons.maxAgeToDays(this.config.maxAge);
+    return Promise.all([
+        this.fetchConcepts(),
+        this.fetchEncounters()
+    ]).then(() => this.formatOutput());
   }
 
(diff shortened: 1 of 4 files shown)


omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
```diff
@@ -32,48 +35,78 @@ export default class ObsAcrossEncountersController {
...
     encounterTypes.push(...legacyEncounterTypes);
 
-    const encounterPromises = encounterTypes.map(e =>
-      this.openmrsRest.get("encounter", {
-        patient: this.config.patientUuid,
-        encounterType: e,
-        v: 'custom:(uuid,encounterDatetime,encounterType:(name,description),obs:(id,uuid,value,concept:(id,uuid,name:(display),datatype:(uuid)),groupMembers:(id,uuid,display,value,concept:(id,uuid,name:(display),datatype:(uuid))))',
-        limit: this.config.maxRecords || 4,
-        fromdate: this.widgetsCommons.maxAgeToDate(this.config.maxAge),
-        order: this.order
```
mogoodrich, line 68: Did we lose this order parameter (maybe I'm just missing it) (and I'm not sure what it does)

mogoodrich: Oh, I see, it's the sort order, but it looks like it was hardcoded to "desc" so we just removed it?

omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
```diff
@@ -32,48 +35,78 @@ export default class ObsAcrossEncountersController {
...
       }
-    });
+      this.output.headers.push('coreapps.date');
+      for (const concept of Object.values(this.conceptsMap)) {
+        this.output.headers.push(concept.display);
+      }
+    }
+
+    this.output.rows = [];
+    const sortedEncs = this.simpleEncs.sort((a, b) => a.encounterDatetime - b.encounterDatetime);
```
mogoodrich, line 89: oh. I see, you just sort by datetime here...

mogoodrich (approved):
Thanks so much getting this in a testable format @brandones !  LGTM....

Let's make sure to give the widget a good visual test after merging since there are so many changes.

omod/src/main/web/dashboardwidgets/obsacrossencounters/obsacrossencounters.controller.js
brandones: Yep. And sort was being handled client-side anyway.

PR #390, opened 2021-03-12
RA-1868:Add spacing between City, State, Country, and Postal Code

sherrif10:
https://issues.openmrs.org/browse/RA-1868
cc @dkayiwa kindly review thanks

### DIFF
--- a/omod/src/main/webapp/fragments/patientdashboard/contactInfo.gsp
+++ b/omod/src/main/webapp/fragments/patientdashboard/contactInfo.gsp
@@ -13,7 +13,7 @@
 <div class="contact-info">
     <div>
         <strong>${ ui.message("coreapps.person.address")}: </strong><br />
-        <div class="ml-2">${ ui.format(config.patient.personAddress).replace("\n", "<br />") }</div>
+        <div class="ml-2">${ ui.format(config.patient.personAddress).replace("\n", "\n","\n","\n","<br />") }</div>
     </div>
     <br />
     <div>


brandones:
Did you run this code? I expect it to produce an error in the browser console, and not to work. Please read the docs for the [replace function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace).

Once you have something working, please include a screenshot, since this is a visual change.

sherrif10:
Thanks @brandones  for reviewing this, do we need to use regex here , will share the screen shot

sherrif10:
Hey @brandones , yeah i think you were right, how about using split(",") method, it seems this was logically done [here](https://github.com/openmrs/openmrs-module-coreapps/blob/master/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/PatientHeaderFragmentController.java#L151)

brandones:
@sherrif10 , that's Java code, that's not even Javascript. That's not even the right programming language. Javascript does have a `split` function, but that example code really has nothing to do with this problem, so I don't know what you are thinking you'll do.

The existing code, `.replace("\n", "<br />")`, seems to me like it should work. The fact that it does not work suggests that solving this problem will require trying to figure out why it does not work. What's happening to that `<br />`? Why isn't it rendering as a line break?

Do you know how to use the browser developer tools? The DOM inspector? That will help.

sherrif10:
@brandones  thanks for your quick response, Let me debug that code using the same line of code, however at first it seemed to be as if it breaks from the java code, will try to follow your suggestions and hopefully will share a clean working code thanks

gracepotma:
@sharif to provide an update. Thanks! :)

PR #391, opened 2021-03-13
RA-1832: Collection.isEmpty() should be used to test for emptiness

dexter808:
Issue : https://issues.openmrs.org/browse/RA-1832

* Changed to use .isEmpty() to check for emptiness instead of .size() for Lists

### DIFF
--- a/api/bin/pom.xml
+++ b/api/bin/pom.xml
@@ -1,0 +1,184 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.openmrs.module</groupId>
+		<artifactId>coreapps</artifactId>
+		<version>1.32.0-SNAPSHOT</version>
+	</parent>
+
+	<artifactId>coreapps-api</artifactId>
+	<packaging>jar</packaging>
+	<name>Core Apps Module API</name>
+	<description>API project for CoreApps</description>
+
+	<dependencies>
+	
+		<dependency>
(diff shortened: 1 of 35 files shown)


api/bin/pom.xml
```diff
@@ -0,0 +1,184 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
HerbertYiga, line 1: Thanks  @dexter808 ,do we nee changes here since we are only replacing .size() with isEmpty()

api/bin/src/main/resources/moduleApplicationContext.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
HerbertYiga, line 1: Are these needed?

api/bin/pom.xml
dexter808: No changes were made in this file.

api/bin/src/main/resources/moduleApplicationContext.xml
dexter808: No changes were made here too.

api/bin/pom.xml
dexter808: Hi @HerbertYiga , I did not touch these files, Maybe these changes were made by the IDE I was using. Is it a problem?

HerbertYiga: > I did not touch these files, Maybe these changes were made by the IDE I was using. Is it a problem?

only needed changes are the ones to be merged,can you try dropping all the uneeded files for your changes

dexter808: Hi @HerbertYiga , sorry for the late reply. I will make a new PR with only the changes that are required.

dexter808:
Issues observed in the pull request.

PR #393, opened 2021-03-22
RA-1903: Check the role 'System Developer' to allow enrollment actions.

icrc-loliveira:
## Issue I worked on

see https://issues.openmrs.org/browse/RA-1903

## Checklist: I completed these to help reviewers :)
- [ ] I have **added tests** to cover my changes. (If you refactored
  existing code that was well tested you do not have to add tests)

- [x] I ran `mvn clean package` right before creating this pull request and
  added all formatting changes to my commit.

- [x] All new and existing **tests passed**.

- [x] My pull request is **based on the latest changes** of the master branch.

### DIFF
--- a/omod/src/main/web/dashboardwidgets/programs/programs.controller.js
+++ b/omod/src/main/web/dashboardwidgets/programs/programs.controller.js
@@ -59,6 +59,12 @@ export default class ProgramsController {
             v: 'custom:(privileges:(name))'
         }).then((response) => {
             if (response && response.user && angular.isArray(response.user.privileges)) {
+                if (response.user.roles.name === 'System Developer') {
+                    this.canEnrollInProgram = true;
+                    this.canEditProgram = true;
+                    this.canDeleteProgram = true;
+                    this.canMarkPatientDead = true;
+                }
                 if (response.user.privileges.some( (p) => { return p.name === 'Task: coreapps.enrollInProgram'; })) {
                     this.canEnrollInProgram = true;
                 };
--- a/omod/src/main/web/dashboardwidgets/programs/programs.controller.js
+++ b/omod/src/main/web/dashboardwidgets/programs/programs.controller.js
@@ -59,7 +59,7 @@ export default class ProgramsController {
             v: 'custom:(privileges:(name))'
         }).then((response) => {
             if (response && response.user && angular.isArray(response.user.privileges)) {
-                if (response.user.roles.name === 'System Developer') {
+                if (response.user.roles.some( (p) => { return p.name === 'System Developer'; })) {
                     this.canEnrollInProgram = true;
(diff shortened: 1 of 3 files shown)


mogoodrich:
For what's it's worth, in the PIH EMR implementation (not sure if there's an equivalent in the Ref App) we have a new role, System Admin, which we give all the privileges that start with "task:" or "app:"... this would likely be easier than trying to add in System Developer checks all throughout the code.  Another alternative would be to change the Session Object so that when it turns the privileges for a system developer, it explicitly returns *all* the privileges in the system.  fyi @mks-d

omod/src/main/web/dashboardwidgets/programs/programs.controller.js
```diff
@@ -59,7 +59,9 @@ export default class ProgramsController {
             v: 'custom:(privileges:(name))'
         }).then((response) => {
             if (response && response.user && angular.isArray(response.user.privileges)) {
-                if (response.user.privileges.some( (p) => { return p.name === 'Task: coreapps.enrollInProgram'; })) {
+                if (response.user.roles.some( (p) => { return p.name === 'System Developer'; })) {
```
dkayiwa, line 62: Could we wrap this in a reusable function such that we do not have the role name `System Developer` in more than one place?

mks-d:
@mogoodrich:

> For what's it's worth, in the PIH EMR implementation (not sure if there's an equivalent in the Ref App) we have a new role, System Admin, which we give all the privileges that start with "task:" or "app:"... this would likely be easier than trying to add in System Developer checks all throughout the code.

That's nifty :-)

However I prefer to actually check whether the privilege is granted `OR` the role 'System Developer' is granted as this is the OpenMRS 3.0 approach, see here: https://openmrs.slack.com/archives/CHP5QAE5R/p1616441537057300?thread_ts=1616192977.045600&cid=CHP5QAE5R

>Another alternative would be to change the Session Object so that when it turns the privileges for a system developer, it explicitly returns _all_ the privileges in the system.

Yes this was discussed on the thread '[_SUPERUSER cannot Enroll Patient in Program_](https://talk.openmrs.org/t/superuser-cannot-enroll-patient-in-program/32271?u=mksd)' and we basically decided to go with the solution implemented in the current PR.

mogoodrich:
Okay, if that's the preferred format going forward, fine with me @mks-d , those as @brandones  suggests we probably want a combined utility method for this...

omod/src/main/web/dashboardwidgets/programs/programs.controller.js
icrc-loliveira: Hi Daniel,
I had created a reusable function in the [dashboardwidgetscommons.service.js](omod/src/main/web/dashboardwidgets/dashboardwidgetscommons.service.js)

mogoodrich (commented):
LGTM!

PR #394, opened 2021-03-23
RA-1905: Timezone support through clientTimezone user property.

icrc-jfigueiredo:
#### Ticket:
https://issues.openmrs.org/browse/RA-1905

#### What I have done:
Using the User propriety to save timezone.

### DIFF
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentController.java
@@ -36,8 +36,22 @@ public class RetrospectiveVisitFragmentController {
             stopDate = startDate;
         }
 
-        //If stopDate is today, it cannot be today at 23:59:59, because that would be in the future.
-        stopDate = stopDate.after(new Date()) ? new Date() : stopDate;
+        if (ui.convertTimezones()) {
+            //If stopDate is today, it cannot be today at 23:59:59, because that would be in the future.
+            stopDate = stopDate.after(new Date()) ? new Date() : stopDate;
+        } else {
+            // set the startDate time component to the start of day
+            startDate = new DateTime(startDate).withTime(0,0,0,0).toDate();
+
+            // if stopDate is today, set stopDate to current datetime, otherwise set time component to end of date
+            if (new DateTime().withTime(0,0,0,0).equals(new DateTime(stopDate).withTime(0,0,0,0))) {
+                stopDate = new Date();
+            }
+            else {
+                stopDate = new DateTime(stopDate).withTime(23, 59, 59, 999).toDate();
+            }
+        }
+
 
 
         try {
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentController.java
@@ -36,16 +36,9 @@ public class RetrospectiveVisitFragmentController {
             stopDate = startDate;
         }
 
-        // set the startDate time component to the start of day
-        startDate = new DateTime(startDate).withTime(0,0,0,0).toDate();
+        //If stopDate is today, it cannot be today at 23:59:59, because that would be in the future.
+        stopDate = stopDate.after(new Date()) ? new Date() : stopDate;
 
-        // if stopDate is today, set stopDate to current datetime, otherwise set time component to end of date
-        if (new DateTime().withTime(0,0,0,0).equals(new DateTime(stopDate).withTime(0,0,0,0))) {
-            stopDate = new Date();
-        }
-        else {
-            stopDate = new DateTime(stopDate).withTime(23, 59, 59, 999).toDate();
-        }
 
         try {
             VisitDomainWrapper createdVisit = adtService.createRetrospectiveVisit(patient, location, startDate, stopDate);
@@ -64,9 +57,15 @@ public class RetrospectiveVisitFragmentController {
 
             if (visits != null) {
                 for (VisitDomainWrapper visit : visits) {
-                    simpleVisits.add(SimpleObject.create("startDate", ui.format(new DateTime(visit.getVisit().getStartDatetime()).toDateMidnight().toDate()),
-                            "stopDate", ui.format(new DateTime(visit.getVisit().getStopDatetime()).toDateMidnight().toDate()),
-                            "id", visit.getVisit().getId(), "uuid", visit.getVisit().getUuid()));
+                    if(ui.handleTimeZones()){
+                        simpleVisits.add(SimpleObject.create("startDate", ui.formatDateWithClientTimezone(new DateTime(visit.getVisit().getStartDatetime()).toDate()),
+                                "stopDate", ui.formatDateWithClientTimezone(new DateTime(visit.getVisit().getStopDatetime()).toDate()),
+                                "id", visit.getVisit().getId(), "uuid", visit.getVisit().getUuid()));
+                    }else{
+                        simpleVisits.add(SimpleObject.create("startDate", ui.format(new DateTime(visit.getVisit().getStartDatetime()).toDateMidnight().toDate()),
+                                "stopDate", ui.format(new DateTime(visit.getVisit().getStopDatetime()).toDateMidnight().toDate()),
+                                "id", visit.getVisit().getId(), "uuid", visit.getVisit().getUuid()));
+                    }
                 }
             }
 
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentController.java
@@ -57,7 +57,7 @@ public class RetrospectiveVisitFragmentController {
 
             if (visits != null) {
                 for (VisitDomainWrapper visit : visits) {
-                    if(ui.handleTimeZones()){
+                    if(ui.convertTimezones()){
                         simpleVisits.add(SimpleObject.create("startDate", ui.formatDateWithClientTimezone(new DateTime(visit.getVisit().getStartDatetime()).toDate()),
                                 "stopDate", ui.formatDateWithClientTimezone(new DateTime(visit.getVisit().getStopDatetime()).toDate()),
                                 "id", visit.getVisit().getId(), "uuid", visit.getVisit().getUuid()));
--- a/omod/src/main/webapp/fragments/patientdashboard/visitIncludes.gsp
+++ b/omod/src/main/webapp/fragments/patientdashboard/visitIncludes.gsp
@@ -52,7 +52,7 @@
     </div>
 </div>
 
-<div id="retrospective-visit-creation-dialog" class="dialog ${ui.handleTimeZones() ? 'rfc3339-date' : ''}" style="display: none">
+<div id="retrospective-visit-creation-dialog" class="dialog ${ui.convertTimezones() ? 'rfc3339-date' : ''}" style="display: none">
     <div class="dialog-header">
         <i class="icon-plus"></i>
         <h3>${ ui.message("coreapps.task.createRetrospectiveVisit.label") }</h3>
@@ -67,7 +67,7 @@
                   visitEndTime = patient.patient.deathDate
                 }
             def endDate =editDateFormat.format(visitEndTime);
-            if(ui.handleTimeZones()){
+            if(ui.convertTimezones()){
                     endDate =visitEndTime;
                 }
             %>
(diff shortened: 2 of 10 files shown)


omod/src/main/webapp/resources/scripts/custom/utilsTimezone.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * Returns date translated according to preferred locale when Date Format displays the month as an abbreviation (MMM).
+ * Ex: Date: 02-Jan-2021 Format: DD-MMM-YYYY Locale: fr_FR --> Return: 02-janv.-2021
+ * @param {object} date Input Date
+ * @param {string} format Date Format
+ * @param {string} locale The preferred locale
+ */
+function formatDatetime(date, format, locale) {
+    var defaultFormat = 'DD.MMM.YYYY, HH:mm:ss';
```
dkayiwa, line 9: The same applies here for our constants naming convention.

omod/src/main/webapp/resources/scripts/custom/utilsTimezone.js
```diff
@@ -0,0 +1,42 @@
...
+    try {
+        moment.locale(locale);
+        return moment(date).format(format);
+    } catch (err) {
+        return moment(date).format(defaultFormat);
+    }
+}
+
+function formatDate(date, format, locale) {
+    var defaultFormat = 'YYYY-MMM-DD';
```
dkayiwa, line 22: The same applies here for our constants naming convention.

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentController.java
```diff
@@ -36,16 +36,9 @@ public Object create(@SpringBean("adtService") AdtService adtService,
             stopDate = startDate;
         }
 
-        // set the startDate time component to the start of day
-        startDate = new DateTime(startDate).withTime(0,0,0,0).toDate();
+        //If stopDate is today, it cannot be today at 23:59:59, because that would be in the future.
+        stopDate = stopDate.after(new Date()) ? new Date() : stopDate;
 
-        // if stopDate is today, set stopDate to current datetime, otherwise set time component to end of date
```
dkayiwa, line 42: Aren't you altering the original behavior even when ui.convertTimezones may not be set?

omod/src/main/webapp/fragments/patientdashboard/visitIncludes.gsp
```diff
@@ -61,14 +66,18 @@
                 if (patient.patient.deathDate) {
                   visitEndTime = patient.patient.deathDate
                 }
+            def endDate =editDateFormat.format(visitEndTime);
```
dkayiwa, line 69: Do you mind putting space after the equal sign?

omod/src/main/webapp/fragments/patientdashboard/visitIncludes.gsp
```diff
@@ -61,14 +66,18 @@
...
+            if(ui.convertTimezones()){
+                    endDate =visitEndTime;
```
dkayiwa, line 71: Same as above.

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentController.java
icrc-jfigueiredo: We changed the way that we format the start/end date, but in the end, everything should be the same as before
Before we sent 2 dates from the client, and then we format on the server side to be startDate 00:00:00 and endDate 23:59:59:59. 
Now we are formatting the 00:00:00 and 23:59:59 on the client side. [here](https://github.com/openmrs/openmrs-module-coreapps/pull/394/files/3adf321846668e81a3048ac3711f8f97e7c72d27#diff-3b4bdfce9ba945b655e5733148593aef49f21e4e31c59f063115c92015d24555R59-R60), this way, if we are using timezone, we can send the date correctly formatted with timezone, if we are not using timezones, we will send the information without timezone, [here](https://github.com/openmrs/openmrs-module-coreapps/pull/394/files/3adf321846668e81a3048ac3711f8f97e7c72d27#diff-3b4bdfce9ba945b655e5733148593aef49f21e4e31c59f063115c92015d24555R66-R68)

dkayiwa: Are the date formatting changes related to time zones?

omod/src/main/webapp/fragments/patientdashboard/visits.gsp
```diff
@@ -76,11 +76,25 @@
         <li class="menu-item viewVisitDetails" data-visit-id="${wrapper.visit.visitId}">
             <span class="menu-date">
                 <i class="icon-time"></i>
-                ${ui.format(wrapper.startDate)}
+                <span class="visit-start-date">
+                    <% if(ui.convertTimezones()) { %>
```
dkayiwa, line 80: Could you remind me as to why we are formatting a different field when convertTimezones = true?

omod/src/main/webapp/pages/markPatientDead.gsp
```diff
@@ -58,6 +58,10 @@
                     jq("#cause-of-death-container > .field-error").append("${ui.message("coreapps.markPatientDead.causeOfDeath.errorMessage")}").show();
                     hasError = true;
                 }
+                if(!hasError && ${ui.convertTimezones()}){
+                    var inputDate = new Date(jq("#death-date-field").val());
```
dkayiwa, line 62: Is the `death-date-field` relevant only when dealing with time zones?

omod/src/main/webapp/resources/scripts/custom/utilsTimezone.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * Returns date translated according to preferred locale when Date Format displays the month as an abbreviation (MMM).
+ * Ex: Date: 02-Jan-2021 Format: DD-MMM-YYYY Locale: fr_FR --> Return: 02-janv.-2021
+ * @param {object} date Input Date
+ * @param {string} format Date Format
+ * @param {string} locale The preferred locale
+ */
+function formatDatetime(date, format, locale) {
+    var DEFAULT_FORMAT = 'DD.MMM.YYYY, HH:mm:ss';
```
dkayiwa, line 9: Is this a constant or variable?

omod/src/main/webapp/resources/scripts/custom/utilsTimezone.js
```diff
@@ -0,0 +1,42 @@
...
+    try {
+        moment.locale(locale);
+        return moment(date).format(format);
+    } catch (err) {
+        return moment(date).format(DEFAULT_FORMAT);
+    }
+}
+
+function formatDate(date, format, locale) {
+    var DEFAULT_FORMAT = 'YYYY-MMM-DD';
```
dkayiwa, line 22: Same as above.

omod/src/main/webapp/resources/scripts/custom/visits.js
```diff
@@ -54,6 +54,19 @@ visit.createRetrospectiveVisitDialog = function(patientId) {
         selector: '#retrospective-visit-creation-dialog',
         actions: {
             confirm: function() {
+                var startVal= jq("#retrospectiveVisitStartDate-field").val();
+                var stopVal=jq("#retrospectiveVisitStopDate-field").val();
```
dkayiwa, line 58: Can we put spaces around the equal signs?

omod/src/test/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentControllerTest.java
```diff
@@ -65,14 +65,9 @@ public void test_shouldCreateNewRetrospectiveVisit() throws Exception {
         Date startDate = new DateTime(2012, 1, 1, 12, 12, 12).toDate();
         Date stopDate = new DateTime(2012, 1, 2, 13,13, 13).toDate();
 
-        // should round to the time components to the start and end of the days, respectively
-        Date expectedStartDate = new DateTime(2012, 1, 1, 0, 0, 0, 0).toDate();
```
dkayiwa, line 69: Because we want to preserve existing behavior when not using timezones, shouldn't we leave existing tests intact and only create new ones to deal with cases where the timezone global property is set?

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentController.java
icrc-jfigueiredo: Yes, when we use timezones, we send datetime and timezone, when we dont use tz we only send datetime

omod/src/main/webapp/fragments/patientdashboard/visits.gsp
icrc-jfigueiredo: In this case its related to using `startDate ` or `startDatetime` when we use TZ we need the datetime information. if we dont use TZ, we only use date information ( this is the current behavior, only date)

omod/src/main/webapp/pages/markPatientDead.gsp
icrc-jfigueiredo: It's important,  because if we are using TZ, we will convert that date to ISO format, [here](https://github.com/openmrs/openmrs-module-coreapps/pull/394/files/51b5742ddba8920005251b5200a9775ee495e384#diff-44221b510d4144f84a4334632ba5928fd06c14facbb588553a1e485117e37ea4R63).  If we are not using TZ we dont need to convert anything...

omod/src/main/webapp/resources/scripts/custom/utilsTimezone.js
icrc-jfigueiredo: Constant, changed the type to const.

omod/src/test/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentControllerTest.java
icrc-jfigueiredo: We changed the way that retrospective visit put the time, as i explained above :

> We changed the way that we format the start/end date, but in the end, everything should be the same as before
> Before we sent 2 dates from the client, and then we format on the server side to be startDate 00:00:00 and endDate 23:59:59:59.
> Now we are formatting the 00:00:00 and 23:59:59 on the client side. here, this way, if we are using timezone, we can send the date correctly formatted with timezone, if we are not using timezones, we will send the information without timezone, here

Because of that, some test need to be adapted, because now, we no longer add the 00:00:00 and 23:59:59 on the server side.

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentController.java
dkayiwa: How about doing the formatting only when the time zone global property is set? That way, we avoid the risk of introducing bugs for those who just want the old behaviour and do not care about time zones.

mogoodrich: I agree with @dkayiwa as I'd prefer not to touch this code if the property is not set, but if it would make things really ugly, I guess I could live with it... we just need to make sure we test fully.

mogoodrich (approved):

omod/src/test/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentControllerTest.java
dkayiwa: If the time zone global property is not set, or set to false, i would expect this test to pass to preserve the old behaviour.

omod/src/test/java/org/openmrs/module/coreapps/fragment/controller/visit/RetrospectiveVisitFragmentControllerTest.java
```diff
@@ -165,7 +166,48 @@ public void test_shouldSetEndDateToCurrentTimeIfEndDateIsCurrentDay() throws Exc
 
     }
 
+    @Test
+    public void test_shouldCreateNewRetrospectiveVisitWithTimezoneAndDontChangeEndDateToEndOfDay() throws Exception {
```
dkayiwa, line 170: These two tests pass even without your above changes. Was your intention to start with failing tests and then make changes which make them pass? In other words, are these tests to cater for the changes in this pull request?

icrc-jfigueiredo: Hello @dkayiwa those 2 tests on my side, only pass with my new changes.

dkayiwa: @icrc-jfigueiredo get a copy of the module without your changes. Then put in only those two tests. You will notices that the both pass on compiling.

PR #395, opened 2021-03-23
Re RA-1832: Collection.isEmpty() should be used to test for emptiness

dexter808:
In response to the previous PR : https://github.com/openmrs/openmrs-module-coreapps/pull/391
Issue : https://issues.openmrs.org/browse/RA-1832

Changed to use .isEmpty() to check for emptiness instead of .size() for Lists


HerbertYiga (approved):
thanks @dexter808

PR #401, opened 2021-04-21
1.31.x

icrc-jfigueiredo:


HerbertYiga:
Thanks @icrc-jfigueiredo ,could you add a description for your changes,ie https://wiki.openmrs.org/display/docs/Pull+Request+Tips

omod/src/main/web/dashboardwidgets/programs/programs.controller.js
```diff
@@ -59,7 +59,9 @@ export default class ProgramsController {
             v: 'custom:(privileges:(name))'
         }).then((response) => {
             if (response && response.user && angular.isArray(response.user.privileges)) {
-                if (response.user.privileges.some( (p) => { return p.name === 'Task: coreapps.enrollInProgram'; })) {
+                if (this.widgetsCommons.isSystemDeveloper(response.user)) {
```
dkayiwa, line 62: How about users who are not system developers but have the `Task: coreapps.enrollInProgram` privilege?

dkayiwa, line 62: Shouldn't we combine this if and else such that we can set `canEnrollInProgram =  true` once instead of twice?

PR #406, opened 2021-05-04
RA-1875: EMPT32 in systemAdministrationPageControllor.java

CrooksDC:
# Description of what I changed
 I check to see if the user doesn't have the privilege and isn't a superuser. If both of these are true then I throw an APIAuthenticationException

# Issue I worked on
an issue where users without the coreapps.systemAdministration privilege are still able to access the system admin page at /openmrs/coreapps/systemadministration/systemAdministration.page

# It can be reproduced following
-Launch OpenMRS application using the URL: http://localhost:8080/openmrs
--
-Provide following info and click login​Username​ as ​admin​ Password​ ​ as ​admin123 Location as ​Inpatient
-Click on “System Administration”.
-Click on “Manage Accounts”.
-Click on the “Add New Account” button.
-In the “Family Name” field, type in “Megan”.
-In the “Given Name” field, type in “Jones”.
-Select Female as Gender.
-Check the “Add User Account?” checkbox.
-In “Username” field type in “mjones”.
-Select “Full” for “Privilege Level”.
-In “Password” and “Confirm Password” fields type in “Megan123*”.
-Uncheck the “Force Password Change” checkbox.
-Click the “Save” button.
-Go to “System administration"
-Copy the URL.
-Log out of the system.
-In the “Username” field type in “mjones”.
-In the “Password” field type in “Megan123*”.
-Select “Registration Desk” as the Location for this Session.
-Click on the “Login” button to log into the system.
-Go to the URL you copied in 17.

# Link to ticket
https://issues.openmrs.org/browse/RA-1875

@isears

### DIFF
--- a/omod/src/main/java/org/openmrs/module/coreapps/page/controller/systemadministration/SystemAdministrationPageController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/page/controller/systemadministration/SystemAdministrationPageController.java
@@ -3,6 +3,7 @@ package org.openmrs.module.coreapps.page.controller.systemadministration;
 import java.util.Collections;
 import java.util.List;
 
+import org.openmrs.api.APIAuthenticationException;
 import org.openmrs.module.appframework.domain.Extension;
 import org.openmrs.module.appframework.service.AppFrameworkService;
 import org.openmrs.module.appui.UiSessionContext;
@@ -17,6 +18,11 @@ public class SystemAdministrationPageController {
                            @SpringBean("appFrameworkService") AppFrameworkService appFrameworkService) {
 
         emrContext.requireAuthentication();
+		
+		if (!emrContext.getCurrentUser().getPrivileges().toString().contains("coreapps.systemAdministration")
+		        && (!emrContext.getCurrentUser().isSuperUser())) {
+			throw new APIAuthenticationException();
+		}
 
         List<Extension> extensions = appFrameworkService.getExtensionsForCurrentUser(SYSTEM_ADMINISTRATION_EXTENSION_POINT);
 
--- a/omod/src/main/java/org/openmrs/module/coreapps/page/controller/systemadministration/SystemAdministrationPageController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/page/controller/systemadministration/SystemAdministrationPageController.java
@@ -17,9 +17,8 @@ public class SystemAdministrationPageController {
     public void controller(PageModel model, UiSessionContext emrContext,
                            @SpringBean("appFrameworkService") AppFrameworkService appFrameworkService) {
 
-        emrContext.requireAuthentication();
-		
-		if (!emrContext.getCurrentUser().getPrivileges().toString().contains("coreapps.systemAdministration")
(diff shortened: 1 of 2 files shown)


omod/src/main/java/org/openmrs/module/coreapps/page/controller/systemadministration/SystemAdministrationPageController.java
```diff
@@ -17,6 +18,11 @@ public void controller(PageModel model, UiSessionContext emrContext,
                            @SpringBean("appFrameworkService") AppFrameworkService appFrameworkService) {
 
         emrContext.requireAuthentication();
+		
+		if (!emrContext.getCurrentUser().getPrivileges().toString().contains("coreapps.systemAdministration")
+		        && (!emrContext.getCurrentUser().isSuperUser())) {
```
HerbertYiga, line 23: Thanks @CrooksDC ,looks like using && operator locks out the other scenarios we want for this,ie we need to make sure that if one doesn't have the system administration privilege's we throw an exception or  incase one is not a super user we throw an exception too, why not use the ||  operator?

CrooksDC: I think it's supposed to be && because you want both of those to be true to throw an exception. If only one is true and you are using the || it will go and throw the expectation even if one is false. Like if it does not have the privilege but it is a superuser and I was using the || it would go into the if statement and throw the exception even though I was a superuser.
@HerbertYiga

HerbertYiga:
> I added a check to see if the user was a superuser or had the privilege to access it. If they were neither I threw an APIAuthenticationException

This kinder conflicts with the above statement you  just made

CrooksDC:
You are right, I need to edit it. I check to see if the user doesn't have the privilege and isn't a superuser. If both of those are true then the exception is thrown. @HerbertYiga

omod/src/main/java/org/openmrs/module/coreapps/page/controller/systemadministration/SystemAdministrationPageController.java
```diff
@@ -17,6 +18,11 @@ public void controller(PageModel model, UiSessionContext emrContext,
                            @SpringBean("appFrameworkService") AppFrameworkService appFrameworkService) {
 
         emrContext.requireAuthentication();
+		
+		if (!emrContext.getCurrentUser().getPrivileges().toString().contains("coreapps.systemAdministration")
```
isears, line 22: What about using `emrContext.getCurrentUser().hasPrivilege("coreapps.systemAdministration")` for the privilege check?

CrooksDC: I just tried to make that change and was unable to get it to work properly using that.

isears: Did you get an error or did it still allow low-privilege users to access the page?

isears (approved):
Great, thanks @CrooksDC !

PR #414, opened 2021-06-21
'Patient Identifier column'

MiguelAHPpih:
As stated last week on teams, we modified the way the patient list works on the main dashboard. Please look into it and tells us what you think about it and if it could work for other countries too. Keep in mind that the way this is suposse to work is for countries to use their national identification documents as patientIdentifier types.

### DIFF
--- a/omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
+++ b/omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
@@ -355,6 +355,19 @@ function PatientSearchWidget(configuration){
                             }
                         });
                     }
+					/* Added Identifier type for the patient list in the main dashboard */
+					else {
+						const identifiersFiltered = patient.identifiers.filter(i => i.identifierType.uuid === "ffbd60fb-599a-4a57-b2f1-4b55847cd938" || i.identifierType.uuid === "3fa0990e-900d-4a80-a4c1-dcf01a0d4f9c");
+						if (identifiersFiltered.length > 0) {
+							const identifierFiltered = identifiersFiltered[0];
+							if (column.type === "identifierType") {
+								columnValue = identifierFiltered.identifierType.name;
+							}
+							else if (column.type === "identifierNumber") {
+								columnValue = identifierFiltered.identifier;
+							}
+						}
+					}
                     if (index === 0) {
                         if(_.contains(initialPatientUuids, patient.uuid)){
                             columnValue += "<span class='recent-lozenge'>" + config.messages.recent + "</span>";


mogoodrich:
Thanks @MiguelAHPpih !  I will take look tomorrow!

omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
```diff
@@ -355,6 +355,19 @@ function PatientSearchWidget(configuration){
                             }
                         });
                     }
+					/* Added Identifier type for the patient list in the main dashboard */
+					else {
+						const identifiersFiltered = patient.identifiers.filter(i => i.identifierType.uuid === "ffbd60fb-599a-4a57-b2f1-4b55847cd938" || i.identifierType.uuid === "3fa0990e-900d-4a80-a4c1-dcf01a0d4f9c");
+						if (identifiersFiltered.length > 0) {
```
mogoodrich, line 361: So, unfortunately, we can't add hardcoded identifier type uuids here because this is a shared OpenMRS module, not specific to PIH.

But I think there's potentially a way forward... if I understand it, we want to have a single column that could show one of two identifier types?

We do currently have functionality to display custom identifier types via the config.  This was implemented earlier this year, see this ticket:

https://issues.openmrs.org/browse/RA-1890

We could create a new ticket about added a feature of allowing multiple identifier types configured for a column, for example, isntead of just supporting this:

```
 {
      "type": "identifier",
      "label": "DNI",
      "value": "${patientIdentifierType.DNI.uuid}",
      "width": "80px"
    },
    {
      "type": "identifier",
      "label": "Passport",
      "value": "${patientIdentifierType.Passport.uuid}",
      "width": "80px"
    },
```

We could support something like this:

```
 {
      "type": "identifier",
      "label": "Identifier",
      "value": "${patientIdentifierType.DNI.uuid},${patientIdentifierType.Passport.uuid}",
      "width": "80px"
    },
```
 I think this would be implemented by modifying the existing code near patientSearchWidget.js:311

Thoughts @MiguelAHPpih  @brandones ?

If we want to take this approach, let's create a ticket in the OpenMRS RA project...

MiguelAHPpih: I spoke with Aldo about this, we agree on this approach.

mogoodrich: Great!

I added a new ticket in OpenMRS Jira for this:
https://issues.openmrs.org/browse/RA-1941

Shall we close this PR and then open a new one labelled with that ticket number once you have an updated commit?

PR #415, opened 2021-06-22
RA-1939 ObsAcrossEncounter Widgets - control datetime format and disp…

icrc-jfigueiredo:
RA-1939 ObsAcrossEncounter Widgets - control datetime format and display encounter provider name

#### Ticket:
https://issues.openmrs.org/browse/RA-1939

#### What I have done:
Add 3 new configs to ObsAcrossEncounter widget:
-  showDateWithTime -> show encounterDate with time componente
-  datetimeFormat -> specify the datetime format
-  showEncounterProviderName -> show the encounter providers

### DIFF
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/dashboardwidgets/DashboardWidgetFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/dashboardwidgets/DashboardWidgetFragmentController.java
@@ -50,6 +50,11 @@ public class DashboardWidgetFragmentController {
         if (appConfig.get("JSDateFormat") == null) {
             appConfig.put("JSDateFormat", adminService.getGlobalProperty(UiFrameworkConstants.GP_FORMATTER_JS_DATE_FORMAT, "YYYY-MMM-DD"));
         }
+
+        if (appConfig.get("JSDateTimeFormat") == null) {
+            appConfig.put("JSDateTimeFormat", adminService.getGlobalProperty(UiFrameworkConstants.GP_FORMATTER_JS_DATETIME_FORMAT, "YYYY-MMM-DD HH:mm"));
+        }
+
         appConfig.put("locale", Context.getLocale().toString());
         appConfig.put("language", Context.getLocale().getLanguage().toString());
 
--- a/omod/src/main/web/dashboardwidgets/dashboardwidgetscommons.service.js
+++ b/omod/src/main/web/dashboardwidgets/dashboardwidgetscommons.service.js
@@ -147,6 +147,23 @@ export default class WidgetsCommons {
     }
 
     /**
+     * Returns date and time translated according to preferred locale.
+     * Ex: Date: 02-Jan-2021 15:10 Format: DD.MMM.YYYY HH:mm Locale: fr --> Return: 02-janv.-2021 15:10
+     *
+     * @param {object} date Input Datetime
+     * @param {string} format Date Format
+     * @param {string} locale The preferred locale
+     */
+    formatDateTime(date, format, locale) {
+        try{
+            moment.locale(locale);
+            return moment(date).format(format);
+        } catch(err) {
+            return moment(date).format("DD.MMM.YYYY HH:mm");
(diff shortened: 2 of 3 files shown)


omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/dashboardwidgets/DashboardWidgetFragmentController.java
```diff
@@ -50,11 +50,16 @@ public void controller(FragmentConfiguration config, @FragmentParam("app") AppDe
         if (appConfig.get("JSDateFormat") == null) {
             appConfig.put("JSDateFormat", adminService.getGlobalProperty(UiFrameworkConstants.GP_FORMATTER_JS_DATE_FORMAT, "YYYY-MMM-DD"));
         }
+
+        if (appConfig.get("JSDateTimeFormat") == null) {
```
dkayiwa, line 54: What happens when the global property is changed at runtime? Since the JSDateTimeFormat in appConfig will not be null, how will once change this value without restarting OpenMRS?

omod/src/main/web/dashboardwidgets/dashboardwidgetscommons.service.js
```diff
@@ -146,6 +146,23 @@ export default class WidgetsCommons {
...
+     * @param {object} date Input Datetime
+     * @param {string} format Date Format
+     * @param {string} locale The preferred locale
+     */
+    formatDateTime(date, format, locale) {
+        try{
+            moment.locale(locale);
+            return moment(date).format(format);
+        } catch(err) {
+            return moment(date).format("DD.MMM.YYYY HH:mm");
```
dkayiwa, line 162: Don't we already have this format in some sort of global property?

icrc-jfigueiredo: Yes we do, that format it's only if the passed format fails, because to use that method we have to choose a format : `formatDateTime(date, format, locale)`

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/dashboardwidgets/DashboardWidgetFragmentController.java
icrc-jfigueiredo: That is a good question, on this case I follow the behavior of `JSDateFormat` and `dateFormat `above. It is suppose to be possible to change that GP on runtime?

dkayiwa: Just go to the settings page and see how you can change a global property at runtime.

icrc-jfigueiredo: I just made some tests it works fine.
So that appConfig.get will check if in the widget configuration we specify some format if we dont specify any, it will try to get the format from the GP, if the global property is null or not present It has a fallback value "YYYY-MMM-DD HH:mm".
So even if we remove the GP it has the fallback value.

dkayiwa: But if we change the global property, at runtime, to a different value, will this value be used without having to restart the application?

icrc-jfigueiredo: Yes it will. If we don't specify the format on the widget config.

dkayiwa: Did you test and confirm it?

icrc-jfigueiredo: Yes, i test it and confirm it.

PR #419, opened 2021-06-29
RA-1942 Redirect to Mark Patient Deceased page when program outcome changed *from* Patient Died

brandones:
https://issues.openmrs.org/browse/RA-1942

Peek 2021-06-29 12-30

### DIFF
--- a/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
+++ b/omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
@@ -390,7 +390,8 @@ export default class ProgramStatusController {
         }).then((response) => {
             // redirect to Mark Patient Dead page if specific "death" outcome is configured and selected
             if (this.markPatientDeadOutcome && this.canMarkPatientDead &&
-                response.outcome && response.outcome.uuid === this.markPatientDeadOutcome) {
+                ((response.outcome && response.outcome.uuid === this.markPatientDeadOutcome)
+                || (this.patientProgram.outcome && this.patientProgram.outcome.uuid == this.markPatientDeadOutcome))) {
               var destinationPage = Handlebars.compile(this.markPatientDeadPage)({
                 patientUuid: this.config.patientUuid,
                 dashboard: this.config.dashboard,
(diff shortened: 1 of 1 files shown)


omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
```diff
@@ -390,7 +390,8 @@ export default class ProgramStatusController {
         }).then((response) => {
             // redirect to Mark Patient Dead page if specific "death" outcome is configured and selected
             if (this.markPatientDeadOutcome && this.canMarkPatientDead &&
-                response.outcome && response.outcome.uuid === this.markPatientDeadOutcome) {
+                ((response.outcome && response.outcome.uuid === this.markPatientDeadOutcome)
+                || (this.patientProgram.outcome && this.patientProgram.outcome.uuid == this.markPatientDeadOutcome))) {
```
mogoodrich, line 394: So I'm thinking this is saying "also redirect if the *existing* outcome was "mark patient died")?   I was thinking we should not redirect in the case where the existing outcome and the set outcome was "mark patient died"... does that handle this case?  

Thinking it through though, I guess it's also arguable that it's better to redirect in that case as well, just to be safe... and it looks like in this case it would redirect with the current death date, which indeed the user might want to be prompted to update in that case...

mogoodrich (commented):
One question... originally I though it needed a change, but may have come around as I wrote the review... lmk your thoughts...

omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
brandones: Yes, it will redirect in any case where either the current status or the new status or both are "dead." I agree that if the patient status is dead and being updated e.g. to died on a different date, that they should be redirected to the death page.

mogoodrich: So, agreed, I like it way you have it... can you update the comment to reflect that it redirects in either of these cases?

mogoodrich (approved):

omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
brandones: Yeah good catch! Done.

PR #423, opened 2021-07-06
RA-1945: ParserEncounterIntoSimpleObjects to cast concepts to ConceptNumeric where applicable.

icrc-jfigueiredo:
**Ticket:**

https://issues.openmrs.org/browse/RA-1945

**What I have done:**

When ParserEncounterIntoSimpleObjects if we are dealing with an concept that is `numeric` we will cast to `conceptNumeric`

### DIFF
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
@@ -136,11 +136,11 @@ public class ParserEncounterIntoSimpleObjects {
             SimpleObject simpleObject = SimpleObject.create("obsId", obs.getObsId());
 
             if (obs.getConcept().isNumeric()) {
-                Boolean integerNumber = false;
+                Boolean isIntegerValue = false;
                 if (!conceptService.getConceptNumeric(obs.getConcept().getConceptId()).isAllowDecimal()) {
-                    integerNumber = true;
+                    isIntegerValue = true;
                 }
-                simpleObject.put("integerNumber", integerNumber);
+                simpleObject.put("isIntegerValue", isIntegerValue);
             }
 
             simpleObject.put("question", capitalizeString(uiUtils.format(obs.getConcept())));
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
@@ -17,6 +17,7 @@ import org.openmrs.Encounter;
 import org.openmrs.Location;
 import org.openmrs.Obs;
 import org.openmrs.Order;
+import org.openmrs.api.ConceptService;
 import org.openmrs.api.LocationService;
 import org.openmrs.module.emrapi.EmrApiProperties;
 import org.openmrs.module.emrapi.diagnosis.Diagnosis;
@@ -44,14 +45,17 @@ public class ParserEncounterIntoSimpleObjects {
 
     private LocationService locationService;
 
+    private ConceptService conceptService;
+
     private DispositionService dispositionService;
 
     public ParserEncounterIntoSimpleObjects(Encounter encounter, UiUtils uiUtils, EmrApiProperties emrApiProperties,
-                                            LocationService locationService, DispositionService dispositionService) {
+                                            LocationService locationService, ConceptService conceptService, DispositionService dispositionService) {
         this.encounter = encounter;
         this.uiUtils = uiUtils;
         this.emrApiProperties = emrApiProperties;
         this.locationService = locationService;
+        this.conceptService = conceptService;
         this.dispositionService = dispositionService;
     }
 	
@@ -130,6 +134,14 @@ public class ParserEncounterIntoSimpleObjects {
         }
         else {
             SimpleObject simpleObject = SimpleObject.create("obsId", obs.getObsId());
+
+            if (obs.getConcept().isNumeric()) {
+                Boolean integerNumber = false;
+                if (!conceptService.getConceptNumeric(obs.getConcept().getConceptId()).isAllowDecimal()) {
+                    integerNumber = true;
+                }
+                simpleObject.put("integerNumber", integerNumber);
+            }
 
             simpleObject.put("question", capitalizeString(uiUtils.format(obs.getConcept())));
             simpleObject.put("answer", uiUtils.format(obs));
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
@@ -19,7 +19,6 @@ import org.openmrs.Obs;
 import org.openmrs.Order;
 import org.openmrs.Concept;
 import org.openmrs.ConceptNumeric;
-import org.openmrs.api.ConceptService;
 import org.openmrs.api.LocationService;
 import org.openmrs.api.db.hibernate.HibernateUtil;
 import org.openmrs.module.emrapi.EmrApiProperties;
@@ -44,8 +43,6 @@ public class ParserEncounterIntoSimpleObjects {
 	
 	private UiUtils uiUtils;
 
-    private ConceptService conceptService;
-
 	private EmrApiProperties emrApiProperties;
 
     private LocationService locationService;
@@ -53,13 +50,12 @@ public class ParserEncounterIntoSimpleObjects {
     private DispositionService dispositionService;
 
     public ParserEncounterIntoSimpleObjects(Encounter encounter, UiUtils uiUtils, EmrApiProperties emrApiProperties,
-                                            LocationService locationService, ConceptService conceptService, DispositionService dispositionService) {
+                                            LocationService locationService, DispositionService dispositionService) {
         this.encounter = encounter;
         this.uiUtils = uiUtils;
         this.emrApiProperties = emrApiProperties;
         this.locationService = locationService;
         this.dispositionService = dispositionService;
-        this.conceptService = conceptService;
     }
 	
 	public List<SimpleObject> parseOrders() {
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
@@ -40,12 +40,12 @@ public class ParserEncounterIntoSimpleObjects {
 	private Encounter encounter;
 	
 	private UiUtils uiUtils;
-	
+
+    private ConceptService conceptService;
+
 	private EmrApiProperties emrApiProperties;
 
     private LocationService locationService;
-
-    private ConceptService conceptService;
 
     private DispositionService dispositionService;
 
@@ -55,8 +55,8 @@ public class ParserEncounterIntoSimpleObjects {
         this.uiUtils = uiUtils;
         this.emrApiProperties = emrApiProperties;
         this.locationService = locationService;
-        this.conceptService = conceptService;
         this.dispositionService = dispositionService;
+        this.conceptService = conceptService;
     }
 	
 	public List<SimpleObject> parseOrders() {
@@ -129,19 +129,14 @@ public class ParserEncounterIntoSimpleObjects {
 	}
 	
 	private SimpleObject parseObs(Obs obs, Locale locale) {
+        if(obs.getConcept() != null && obs.getConcept().isNumeric()){
+            obs.setConcept(this.conceptService.getConceptNumeric(obs.getConcept().getConceptId()));
+        }
         if ("org.openmrs.Location".equals(obs.getComment())) {
             return (parseObsWithLocationAnswer(obs, locationService.getLocation(Integer.valueOf(obs.getValueText()))));
         }
         else {
             SimpleObject simpleObject = SimpleObject.create("obsId", obs.getObsId());
-
-            if (obs.getConcept().isNumeric()) {
-                Boolean isIntegerValue = false;
-                if (!conceptService.getConceptNumeric(obs.getConcept().getConceptId()).isAllowDecimal()) {
-                    isIntegerValue = true;
-                }
-                simpleObject.put("isIntegerValue", isIntegerValue);
-            }
 
             simpleObject.put("question", capitalizeString(uiUtils.format(obs.getConcept())));
             simpleObject.put("answer", uiUtils.format(obs));
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
@@ -128,13 +128,11 @@ public class ParserEncounterIntoSimpleObjects {
 	}
 	
 	private SimpleObject parseObs(Obs obs, Locale locale) {
-        if (obs.getConcept() != null) {
-            //TODO HibernateUtil.getRealObjectFromProxy(obs.getConcept()) should be moved to a better place in the future,
-            //If we remove it, the concept type will never be an instanceof ConceptNumeric
-            Concept concept = HibernateUtil.getRealObjectFromProxy(obs.getConcept());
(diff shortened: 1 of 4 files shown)


omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
```diff
@@ -131,6 +135,14 @@ private SimpleObject parseObs(Obs obs, Locale locale) {
         else {
             SimpleObject simpleObject = SimpleObject.create("obsId", obs.getObsId());
 
+            if (obs.getConcept().isNumeric()) {
+                Boolean integerNumber = false;
```
mks-d, line 139: Why not `allowDecimals`? So you stick to the concept numeric naming.

omod/src/test/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjectsTest.java
```diff
@@ -297,6 +300,29 @@ public void testParsingSimpleObs() throws Exception {
...
+		ConceptNumeric conceptNumeric = new ConceptNumeric();
+		conceptNumeric.setId(1000);
+		conceptNumeric.addName(conceptName);
+		conceptNumeric.setUuid("uuid");
+		conceptNumeric.setAllowDecimal(false);
+		conceptNumeric.setDatatype(numericDatatype);
+
+		Double numericValue = 2.0;
+
+		when(conceptService.getConceptNumeric(anyInt())).thenReturn(conceptNumeric);
```
mks-d, line 319: I think a predicate like `eq(1000)` would be better than `anyInt()`.

omod/src/test/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjectsTest.java
```diff
@@ -297,6 +300,29 @@ public void testParsingSimpleObs() throws Exception {
 		assertThat(path(parsed.getObs(), 0, "answer"), is((Object) valueText));
 	}
 
+	@Test
+	public void testParsingNumericDataTypeWithAllowDecimalAsFalseToSimpleObs() throws Exception {
+		ConceptDatatype numericDatatype  = new ConceptDatatype() ;
+		numericDatatype.setName("Numeric");
+		numericDatatype.setHl7Abbreviation("NM");
```
mks-d, line 307: Do you need this?

omod/src/test/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjectsTest.java
```diff
@@ -297,6 +300,29 @@ public void testParsingSimpleObs() throws Exception {
 		assertThat(path(parsed.getObs(), 0, "answer"), is((Object) valueText));
 	}
 
+	@Test
+	public void testParsingNumericDataTypeWithAllowDecimalAsFalseToSimpleObs() throws Exception {
```
mks-d, line 304: `testParsingNumericObsShouldHandleAllowDecimal`

It would be great to test also the case when allo decimal is `true`.

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
icrc-jfigueiredo: But then I would always need to check the false case... That is the only reason...

omod/src/test/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjectsTest.java
icrc-jfigueiredo: Yes, I need, if I don't put that I will have NPE.

omod/src/test/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjectsTest.java
icrc-jfigueiredo: Sure, I will create another test for that case.

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
icrc-jfigueiredo: Changed to `isIntegerValue`

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
```diff
@@ -131,6 +135,14 @@ private SimpleObject parseObs(Obs obs, Locale locale) {
         else {
             SimpleObject simpleObject = SimpleObject.create("obsId", obs.getObsId());
 
+            if (obs.getConcept().isNumeric()) {
+                Boolean isIntegerValue = false;
+                if (!conceptService.getConceptNumeric(obs.getConcept().getConceptId()).isAllowDecimal()) {
+                    isIntegerValue = true;
+                }
+                simpleObject.put("isIntegerValue", isIntegerValue);
```
dkayiwa, line 143: How about `allowDecimal` instead of `isIntegerValue`?

omod/src/test/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjectsTest.java
```diff
@@ -296,6 +300,51 @@ public void testParsingSimpleObs() throws Exception {
 		assertThat(path(parsed.getObs(), 0, "question"), is((Object) consultNote));
 		assertThat(path(parsed.getObs(), 0, "answer"), is((Object) valueText));
 	}
+	@Test
+	public void testParsingNumericObsShouldHandleAllowDecimal() throws Exception {
+		ConceptDatatype numericDatatype  = new ConceptDatatype() ;
```
dkayiwa, line 305: These two tests pass even without your above fix. I thought they would start by failing and then your changes above make them pass. Correct?

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
icrc-jfigueiredo: The only reason that I use `isIntegerValue` it's because this way I can evaluate the `true` instead of the false case because we want the case when we cannot use the decimals...

omod/src/test/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjectsTest.java
icrc-jfigueiredo: Yes, i made some adjustments

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
```diff
@@ -125,6 +129,9 @@ public int compare(SimpleObject o1, SimpleObject o2) {
 	}
 	
 	private SimpleObject parseObs(Obs obs, Locale locale) {
+        if(obs.getConcept() != null && obs.getConcept().isNumeric()){
+            obs.setConcept(this.conceptService.getConceptNumeric(obs.getConcept().getConceptId()));
+        }
```
mks-d, line 134: @mogoodrich this is the actual change, we make sure that concepts are propagated as `ConceptNumeric` in the obs representations. This helps kick in some concept numeric specific logic down the line.

mks-d (approved):

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
mogoodrich: That's bizarre... so on Obs the concept is of class Concept even when it's numeric, so we reload the concept to cast it back to ConceptNumeric?

mks-d: @mogoodrich indeed. Actually if we had the time, I wanted @JFFigueiredo to investigate a little bit where/why this cast does not happen already up the chain.

It could be that this is because we access the obs through the encounter (so through `EncounterService` and not through `ObsService`).

@Ruhanga maybe you could spend 30 min on this to clarify what happens? As in:
* Write a unit test (or debug an existing unit test) that fetches an obs numeric via its encompassing encounter, and check whether the obs' concept is cast or not into `ConceptNumeric`.

Ruhanga: Ok...

mogoodrich: If we don't figure it out and merge this change as-is, let's add a comment to the code saying we aren't quite sure why we have to do this? :)

Ruhanga: I've tried to reproduce the issue through [this commit](https://github.com/openmrs/openmrs-core/commit/b2794c526565f7cbe1ed0983c6ae9ac1cb3df7dd) but seem not to get exactly to problem for now. I'll take time over the weekend to investigate further.

mks-d: @mogoodrich I have asked @JFFigueiredo to give simple steps to reproduce the issue on QA Ref App.

In short, if a concept numeric disallows decimals, then everything should work already and 'Show Details' should not display any decimals. However this isn't the case, precisely because of the `ConceptNumeric` casting not happening issue.

If everything takes too much time to investigate, we will merge the current PR with a "no idea why we have to do this" comment as you suggested :-)

icrc-jfigueiredo: @mks-d , @mogoodrich I added the steps to reproduce the bug on this ticket [RA-1945](https://issues.openmrs.org/browse/RA-1945).
Here is a quick view of the problem, even when allowDecimals not enabled, it add the ".0" on "Show Details"
integer

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
```diff
@@ -125,6 +129,9 @@ public int compare(SimpleObject o1, SimpleObject o2) {
 	}
 	
 	private SimpleObject parseObs(Obs obs, Locale locale) {
+        if(obs.getConcept() != null && obs.getConcept().isNumeric()){
```
dkayiwa, line 132: According to [this](https://github.com/openmrs/openmrs-core/blob/2.4.0/api/src/main/java/org/openmrs/Concept.java#L1337), `obs.getConcept().isNumeric()` will always return `false` when the return value of getConcept() is of class `Concept`.
And according to [this](https://github.com/openmrs/openmrs-core/blob/2.4.0/api/src/main/java/org/openmrs/ConceptNumeric.java#L196), the only time that `obs.getConcept().isNumeric()` will return `true` is when the return value of getConcept() is of class `ConceptNumeric` and the data type is `Numeric`.
So i do not see the value of the next line of code below.

icrc-jfigueiredo: Hello @dkayiwa I understand that but can you see this example: 
instance_of_numeric

icrc-jfigueiredo: When i stop at a breakpoint, the `obs.getConcept().isNumeric()` is `true` but `obs.getConcept() instanceof ConceptNumeric` is `false`.

dkayiwa: Which OpenMRS platform version are you running?

icrc-jfigueiredo: I am using `<openMRSVersion>2.3.3-SNAPSHOT</openMRSVersion>`.

dkayiwa: Instead of using a snapshot version of the platform, do you mind trying a released tag like `2.3.3` and see if you can still reproduce the problem?

icrc-jfigueiredo: With version `2.3.3` nothing change, I can still reproduce the problem.

dkayiwa: > but `obs.getConcept() instanceof ConceptNumeric` is `false`.

Then it must be a hibernate proxy. Does `HibernateUtil.getRealObjectFromProxy()` help?

icrc-jfigueiredo: @dkayiwa if i do `HibernateUtil.getRealObjectFromProxy(obs.getConcept()) instanceof ConceptNumeric` it is true. So the problem should be when i get the concept via Hibernate ( maybe lazy load) ?

dkayiwa: Assign the concept property to the return value of HibernateUtil.getRealObjectFromProxy

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
```diff
@@ -125,6 +128,12 @@ public int compare(SimpleObject o1, SimpleObject o2) {
 	}
 	
 	private SimpleObject parseObs(Obs obs, Locale locale) {
```
dkayiwa, line 130: Concept concept = obs.getConcept();?

icrc-jfigueiredo: If i use `Concept concept = obs.getConcept();` instead of  `Concept concept = HibernateUtil.getRealObjectFromProxy(obs.getConcept());` the concept will not be a conceptNumeric.

icrc-jfigueiredo: Good idea.

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/ParserEncounterIntoSimpleObjects.java
```diff
@@ -125,13 +128,19 @@ public int compare(SimpleObject o1, SimpleObject o2) {
...
+                obs.setConcept(HibernateUtil.getRealObjectFromProxy(concept));
+        }
         if ("org.openmrs.Location".equals(obs.getComment())) {
             return (parseObsWithLocationAnswer(obs, locationService.getLocation(Integer.valueOf(obs.getValueText()))));
         }
         else {
             SimpleObject simpleObject = SimpleObject.create("obsId", obs.getObsId());
 
-            simpleObject.put("question", capitalizeString(uiUtils.format(obs.getConcept())));
+            simpleObject.put("question", capitalizeString(uiUtils.format(concept)));
```
dkayiwa, line 143: Isn't `concept` still a hibernate proxy?

icrc-jfigueiredo: Yes, sorry, it can be `obs.getConcept()` to use the value with `HibernateUtil`

PR #429, opened 2021-09-09
RA-1954 - Add GP DECEASED_DATE_USING_TIME 

icrc-jfigueiredo:
Issue: https://issues.openmrs.org/browse/RA-1954

Create a global property `DECEASED_DATE_USING_TIME` that allows choosing if we want the Decease Date datepicker to have a time component.

If it's true we can select date + time.

If it's not true, we can only select date.

### DIFF
--- a/api/src/main/java/org/openmrs/module/coreapps/CoreAppsConstants.java
+++ b/api/src/main/java/org/openmrs/module/coreapps/CoreAppsConstants.java
@@ -58,4 +58,6 @@ public class CoreAppsConstants {
    public static final String GLOBAL_PROPERTY_CONDITIONS_CRITERIA = "coreapps.conditionListClasses";
 
    public static final String DEFAULT_CODING_SOURCE = "ICD-10-WHO";
+
+   public static final String DECEASED_DATE_USING_TIME = "coreapps.deceasedDateUsingTime";
 }
--- a/api/src/main/java/org/openmrs/module/coreapps/CoreAppsConstants.java
+++ b/api/src/main/java/org/openmrs/module/coreapps/CoreAppsConstants.java
@@ -59,5 +59,5 @@ public class CoreAppsConstants {
 
    public static final String DEFAULT_CODING_SOURCE = "ICD-10-WHO";
 
-   public static final String DECEASED_DATE_USING_TIME = "coreapps.deceasedDateUsingTime";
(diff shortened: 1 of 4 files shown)


mogoodrich (commented):
Looks like this is a  draft, but LGTM so far, thanks @icrc-jfigueiredo

api/src/main/java/org/openmrs/module/coreapps/CoreAppsConstants.java
```diff
@@ -58,4 +58,6 @@
    public static final String GLOBAL_PROPERTY_CONDITIONS_CRITERIA = "coreapps.conditionListClasses";
 
    public static final String DEFAULT_CODING_SOURCE = "ICD-10-WHO";
+
+   public static final String DECEASED_DATE_USING_TIME = "coreapps.deceasedDateUsingTime";
```
dkayiwa, line 62: Can you use the same naming convention like the rest of the global property constants in this class?

omod/src/main/webapp/pages/markPatientDead.gsp
```diff
@@ -127,7 +127,7 @@ ${ui.includeFragment("coreapps", "patientHeader", [patient: patient])}
                         formFieldName: "deathDate",
                         left         : true,
                         defaultDate  : patient?.getDeathDate() ?: defaultDeathDate ?: null,
-                        useTime      : false,
+                        useTime      : deceasedDateTimeComponent,
```
dkayiwa, line 130: The name `deceasedDateTimeComponent` does not look like a boolean variable. It sounds like an actual component.

icrc-jfigueiredo: What do you think of `isDeceasedDateTime` ?

dkayiwa: How about something like? `includesTime`

api/src/main/java/org/openmrs/module/coreapps/CoreAppsConstants.java
icrc-jfigueiredo: Sure i will change it.

omod/src/main/webapp/pages/markPatientDead.gsp
icrc-jfigueiredo: I will change to that.

PR #431, opened 2021-09-21
RA-1960, order obs by datetime descending

cioan:
Just ordering the obs by encounter datetime in the descending order. 

https://issues.openmrs.org/browse/RA-1960


brandones (approved):
Dunno where this requirement came from, but seems reasonable enough, and LGTM!

mogoodrich (commented):
Let's revert this and implement this as a configuration parameter (see my comment on the ticket) since we don't want to change the behavior of a widget that is currently used by many... fyi @cioan  @brandones

PR #433, opened 2021-09-23
RA-1961 Date Of Birth converting to a different timezone

icrc-jfigueiredo:
Issue: [RA-1961](https://issues.openmrs.org/browse/RA-1961)

Change Birth Date to use ui.formatDateWithoutTimezoneConversion if we are using timezones, to make sure we don't have any type of conversion on that date.


omod/src/main/webapp/fragments/patientHeader.gsp
```diff
@@ -113,7 +113,13 @@
                     <% } else { %>
                         ${ui.message("coreapps.ageDays", patient.ageInDays)}
                     <% } %>
-                    (<% if (patient.birthdateEstimated) { %>~<% } %>${ ui.formatDatePretty(patient.birthdate) })
+                    (<% if (patient.birthdateEstimated) { %>~<% } %>
```
HerbertYiga, line 116: thanks @icrc-jfigueiredo  cc  @dkayiwa

dkayiwa: @icrc-jfigueiredo did you test these changes by running the module and see the user interface effects?

icrc-jfigueiredo: Yes, I did test it, with Timzone GP as `true` and `false`, and it works well.

dkayiwa: @icrc-jfigueiredo did you test it before or after removing the extra `formatDatePretty` that you had below?

icrc-jfigueiredo: After removing it.

icrc-jfigueiredo: But I think we should release a new version of UiFramework, because we are using a method that is only available on the snapshot version.

PR #442, opened 2021-11-29
RA-1979 Program enrollment widget stack traces with NPE

mseaton:


mogoodrich (approved):
LGTM, thanks @mseaton

PR #445, opened 2022-02-23
#RA-1983 - Refresh component after data is fetched

icrc-toliveira:
The issue (https://issues.openmrs.org/browse/RA-1983):

ObsAcrossEncounters widgets might not show information or show only a small fraction

How to replicate:
1) Go to a page containing multiple ObsAcrossEncounters widgets (These should contain information)
2) Go into the browser dev tools -> Network tab -> make sure "Disable cache" is unchecked 
3) Refresh the page

Solution:
Add scope to component 
And refresh scope after all data has been fetched


cioan (approved):
LGTM! Thanks @icrc-toliveira !

PR #450, opened 2022-03-15
RA-1985 - Allow changing visit time dynamically

icrc-toliveira:
Issue: https://issues.openmrs.org/browse/RA-1985

If a user creates a visit, and later on decide to edit said visit.
Currently its no possible to change time.  

The proposal is allowing the visit time to be changed
(According with the needs of the implementation)

Solution found: 
Adding global property to enable / disable useTime in the dateTimePicker

### DIFF
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/VisitDatesFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/VisitDatesFragmentController.java
@@ -13,6 +13,7 @@
  */
 package org.openmrs.module.coreapps.fragment.controller.visit;
 
+import org.apache.commons.lang.BooleanUtils;
 import org.joda.time.DateTime;
 import org.openmrs.Visit;
 import org.openmrs.api.VisitService;
@@ -39,7 +40,7 @@ public class VisitDatesFragmentController {
                                             @RequestParam(value="stopDate", required = false) Date stopDate,
                                             HttpServletRequest request, UiUtils ui) {
 
-        Boolean allowChangingVisitTime = new Boolean(Context.getAdministrationService().getGlobalProperty(CoreAppsConstants.GP_ALLOW_CHANGING_VISIT_TIME));
+        boolean allowChangingVisitTime = BooleanUtils.toBoolean(Context.getAdministrationService().getGlobalProperty(CoreAppsConstants.GP_ALLOW_CHANGING_VISIT_TIME));
 
         if (!allowChangingVisitTime && !isSameDay(startDate, visit.getStartDatetime())) {
             visit.setStartDatetime(new DateTime(startDate).toDateMidnight().toDate());
--- a/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/VisitDatesFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/VisitDatesFragmentController.java
@@ -44,7 +44,7 @@ public class VisitDatesFragmentController {
 
         if (!allowChangingVisitTime && !isSameDay(startDate, visit.getStartDatetime())) {
             visit.setStartDatetime(new DateTime(startDate).toDateMidnight().toDate());
-        }else{
+        } else if (allowChangingVisitTime && !isSameDay(startDate, visit.getStartDatetime())){
             visit.setStartDatetime(new DateTime(startDate).toDate());
         }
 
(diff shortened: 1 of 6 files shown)


omod/src/main/resources/config.xml
```diff
@@ -302,6 +302,14 @@
         </description>
     </globalProperty>
 
+    <globalProperty>
+        <property>coreapps.allowChangingVisitTime</property>
+        <defaultValue>true</defaultValue>
```
dkayiwa, line 307: Shouldn't this default to false in order to preserve the existing behaviour?

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/VisitDatesFragmentController.java
```diff
@@ -37,9 +39,12 @@ public SimpleObject setDuration(@SpringBean("visitService") VisitService visitSe
                                             @RequestParam(value="stopDate", required = false) Date stopDate,
                                             HttpServletRequest request, UiUtils ui) {
 
+        Boolean allowChangingVisitTime = new Boolean(Context.getAdministrationService().getGlobalProperty(CoreAppsConstants.GP_ALLOW_CHANGING_VISIT_TIME));
```
dkayiwa, line 42: How about `org.apache.commons.lang.BooleanUtils.toBoolean()`?

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/VisitDatesFragmentController.java
```diff
@@ -37,9 +39,12 @@ public SimpleObject setDuration(@SpringBean("visitService") VisitService visitSe
...
 
-        if (!isSameDay(startDate, visit.getStartDatetime())) {
+        if (!allowChangingVisitTime && !isSameDay(startDate, visit.getStartDatetime())) {
             visit.setStartDatetime(new DateTime(startDate).toDateMidnight().toDate());
+        }else{
+            visit.setStartDatetime(new DateTime(startDate).toDate());
```
dkayiwa, line 47: Do you still want this to be executed if `allowChangingVisitTime = false` but `isSameDay(startDate, visit.getStartDatetime()) == true`?

omod/src/main/resources/config.xml
mogoodrich: Agreed!

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/VisitDatesFragmentController.java
mogoodrich: I'll let @icrc-toliveira  answer but I do think this is the correct behavior... we only want to "reset" the time component to midnight if allow change visit time is false and we are changing the date... if allowChangeVisitTime is true, we always want to set the visit time to the time passed in via the submit.

mogoodrich (changes requested):
Looks good... we definitely do want to set the GP to false by default, but otherwise I think this is good to go.

omod/src/main/resources/config.xml
icrc-toliveira: Done

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/VisitDatesFragmentController.java
icrc-toliveira: Hi  @dkayiwa and @mogoodrich 

I just did a small update, this should account for @dkayiwa concerns, while keeping the behavior " reset the time to midnight if allow change visit time is false and we are changing the date"

dkayiwa: @icrc-toliveira i find that this logic would be easier to understand if you put the `allowChangingVisitTime` below `if (!isSameDay(startDate, visit.getStartDatetime())) {`

That way, you do not have to repeat `!isSameDay(startDate, visit.getStartDatetime())`

omod/src/main/resources/config.xml
```diff
@@ -302,6 +302,14 @@
         </description>
     </globalProperty>
 
+    <globalProperty>
+        <property>coreapps.allowChangingVisitTime</property>
+        <defaultValue>false</defaultValue>
+        <description>
+            Be able to change the start hour of a visit.
```
dkayiwa, line 309: Only hour, not minute too? Oh would simply say change the start time?

icrc-toliveira: Done

omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/visit/VisitDatesFragmentController.java
icrc-toliveira: Done

PR #454, opened 2022-03-16
RA-1986: CoreApps Widget Commons: "daysSinceDate" and "dateToDaysAgo"…

mogoodrich:
… fail due to daylight savings


mogoodrich:
see: https://issues.openmrs.org/browse/RA-1986

PR #458, opened 2022-11-03
Patient Search:Add ability to add column that shows patient identifie…

louidorjp:
please review this PR @mogoodrich

### DIFF
--- a/omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
+++ b/omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
@@ -365,9 +365,15 @@ function PatientSearchWidget(configuration){
                     }
                     else if (column.type === 'identifierLocation') {
                         jq.each(patient.identifiers, function (index, patientIdentifier) {
+                            var identifierType = patientIdentifier.identifierType;
                             var identifierLocation = patientIdentifier.location;
-                            if (identifierLocation != null) {
-                                columnValue = identifierLocation.display
+                            if (identifierType != null && !patientIdentifier.voided &&
+                                    column.value && column.value.includes(identifierType.uuid)) {
+                                    if (identifierLocation != null) {
+                                        columnValue = columnValue
+                                          + (columnValue !== '' ? ', ' : '') 
+                                          + identifierLocation.display;
+                                    }
                             }
                         });
                     }
--- a/omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
+++ b/omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
@@ -170,7 +170,7 @@ function PatientSearchWidget(configuration){
     var customRep = 'custom:(patientId,uuid,' +
                     'patientIdentifier:(uuid,identifier),' +
                     'person:(gender,age,birthdate,birthdateEstimated,personName),' +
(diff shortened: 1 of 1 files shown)


omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
```diff
@@ -363,6 +363,14 @@ function PatientSearchWidget(configuration){
                             }
                         });
                     }
+                    else if (column.type === 'location') {
```
mogoodrich, line 366: let's call this 'identifierLocation' instead of 'location' to make it clear that it's the location of the identifier that we are displaying.

omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
```diff
@@ -363,6 +363,14 @@ function PatientSearchWidget(configuration){
...
+                        jq.each(patient.identifiers, function(index, identifier) {
```
mogoodrich, line 367: Instead of just looping through all the identifiers and picking the first one, let's set it so it is configured to match a specific identifier type by uuid, like how the identifier functionality works.

for example, we'd want to be able set a value to specific the identifier type like we do when displaying an identifier:

https://github.com/PIH/openmrs-config-zl/blob/master/configuration/pih/pih-config-haiti-hiv.json#L81 (where in this case "139766e8-15f5-102d-96e4-000c29c2a5d7" is the uuid column of the HIVEMR-V1 in the patient_identifier_type table)

Code would likely be similar to the code in the link below, although I think it can be simpler, because I don't think we need to support asking for multiple locations, we can assume just a single uuid and skip all the logic around "isMultipleIdentifierTypes":

https://github.com/openmrs/openmrs-module-coreapps/blob/master/omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js#L313

omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
```diff
@@ -363,6 +363,14 @@ function PatientSearchWidget(configuration){
                             }
                         });
                     }
+                    else if (column.type === 'identifierLocation') {
+                        jq.each(patient.identifiers, function (index, patientIdentifier) {
```
mogoodrich, line 367: (let's add the part that matched on identifier type)

var patientIdentifierType = patientIdentifier.identifierType;
// find all non-voided patientIdentifiers of whose type matches that defined in column.value
if (!patientIdentifier.voided && patientIdentifierType != null && column.value && column.value===patientIdentifierType.uuid) {

omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
```diff
@@ -363,6 +363,14 @@ function PatientSearchWidget(configuration){
...
+                            var identifierLocation = patientIdentifier.location;
+                            if (identifierLocation != null) {
+                                columnValue = identifierLocation.display
```
mogoodrich, line 370: change to:  (the part in the in middle makes sure we add a comma between each value)

// create a string of all matching identifier values
columnValue = columnValue + (columnValue !== '' ? ', ' : '') + identifierLocation.display

mogoodrich (approved):
Thanks @louidorjp , merging now!

PR #460, opened 2022-12-20
RA-1998

mogoodrich:


omod/src/main/web/dashboardwidgets/programstatus/programstatus.controller.js
```diff
@@ -738,22 +735,9 @@ export default class ProgramStatusController {
         this.$window.location.reload();
     }
 
-    dateWithoutTime(date) {
-        return moment(date).format('YYYY-MM-DD');
-    }
-
-    toUTCDate(dateString) {
```
mogoodrich, line 745: Note that this toUTCDate function was added to fix a similar issue to this issue (see RA-1774) however, as far as I can tell, the assumption this function was making was that the server dates are always in UTC, which may not be the case)

mogoodrich:
Updated PR ready for review, for description see: https://issues.openmrs.org/browse/RA-1998

mseaton (approved):
All this looks well done to me @mogoodrich .  It would be good to get some input from those who added the original handling of this feature, to provide some additional review and (ideally) testing, once merged.

mogoodrich:
Very good point @mseaton ... I had mean to flag @mks-d  but forgot and added him now... @mks-d  I believe (but I could be wrong) that the initial fix was applied for the Mekom use case.  Is there anyone at Mekom that can review and/or test once it's merged in.

mks-d (commented):
@mogoodrich please go ahead. I don't think this widget is still currently used on our end. Thanks for the heads up.

mogoodrich:
Thanks all, merging!

PR #466, opened 2023-03-21
RA-1678 Condition List: Problems with time zones

mseaton:
This PR addresses a number of issues with the Condition List date fields:

Reported issues fixed:
* It fixes RA-1678 such that dates submitted to the server are stored consistently with how they are entered on the client.
* It fixes RA-2012 by enabling the onset and end date input fields to be cleared/unset

Unreported issues fixed:
* Fixes syntax errors in JS causing Jasmine tests to fail
* Fixes bug that prevents marking conditions as inactive if the locale is anything other than English
* Fixes the display of onset and end dates in the condition list table to match refapp localized date formatting.

### DIFF
--- a/omod/src/main/webapp/pages/conditionlist/manageCondition.gsp
+++ b/omod/src/main/webapp/pages/conditionlist/manageCondition.gsp
@@ -60,7 +60,8 @@ ${ui.includeFragment("coreapps", "patientHeader", [patient: patient])}
                         datePickerFormat : "yyyy-MM-dd",
                         label            : ui.message('coreapps.conditionui.onsetdate'),
                         useTime          : false,
-                        endDate          : new Date()
+                        endDate          : new Date(),
+                        clearButton      : true
                 ])}
             </li>
             &nbsp;&nbsp;
@@ -70,17 +71,18 @@ ${ui.includeFragment("coreapps", "patientHeader", [patient: patient])}
                         datePickerFormat : "yyyy-MM-dd",
                         label            : ui.message('coreapps.stopDate.label'),
                         useTime          : false,
-                        endDate          : new Date()
+                        endDate          : new Date(),
+                        clearButton      : true
                 ])}
             </li>
             <br/><br/>
             <div id="status" class="horizontal">
                 <p>
-                    <input type="radio" id="status-1" class="condition-status" value="${ui.message('coreapps.conditionui.active.label')}" name="status" ng-model="condition.status"  ng-change="showEndDate()"/>
+                    <input type="radio" id="status-1" class="condition-status" value="ACTIVE" name="status" ng-model="condition.status"  ng-change="showEndDate()"/>
                     <label for="status-1">${ui.message('coreapps.conditionui.active.label')}</label>
                 </p>
                 <p>
-                    <input type="radio" id="status-2" class="condition-status" value="${ui.message('coreapps.conditionui.inactive.label')}" name="status" ng-model="condition.status"  ng-change="showEndDate()"/>
+                    <input type="radio" id="status-2" class="condition-status" value="INACTIVE" name="status" ng-model="condition.status"  ng-change="showEndDate()"/>
                     <label for="status-2">${ui.message('coreapps.conditionui.inactive.label')}</label>
                 </p>
             </div>
--- a/omod/src/main/webapp/pages/conditionlist/manageConditions.gsp
+++ b/omod/src/main/webapp/pages/conditionlist/manageConditions.gsp
@@ -6,6 +6,7 @@
     ui.includeJavascript("uicommons", "angular-resource.min.js")
     ui.includeJavascript("uicommons", "angular-common.js")
     ui.includeJavascript("uicommons", "underscore-min.js")
+    ui.includeJavascript("uicommons", "moment-with-locales.min.js")
 
     ui.includeJavascript("coreapps", "conditionlist/lib/polyfills.js")
     ui.includeJavascript("coreapps", "conditionlist/restful-services/restful-service.js");
@@ -40,7 +41,7 @@ ${ui.includeFragment("coreapps", "patientHeader", [patient: patient])}
 <h2>${ui.message("coreapps.conditionui.manageConditions")}</h2>
 
 <div id="condition" ng-app="conditionsApp" ng-controller="ConditionsController"
-     ng-init="conditions = getConditions('${patient.uuid}')">
+     ng-init="conditions = getConditions('${patient.uuid}'); config = { dateFormat: '${ ui.getJSDateFormat() }', locale: '${ ui.getLocale().getLanguage() }'};">
     <div id="tabs">
         <ul>
             <li>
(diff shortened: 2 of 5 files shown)


omod/src/main/webapp/pages/conditionlist/manageCondition.gsp
```diff
@@ -70,17 +71,18 @@ ${ui.includeFragment("coreapps", "patientHeader", [patient: patient])}
                         datePickerFormat : "yyyy-MM-dd",
                         label            : ui.message('coreapps.stopDate.label'),
                         useTime          : false,
-                        endDate          : new Date()
+                        endDate          : new Date(),
+                        clearButton      : true
```
mseaton, line 75: The datetimepicker in uicommons has a boolean configuration property to indicate whether a clear button should be included in the widget.  This ended up being all that was needed to fix RA-2012

omod/src/main/webapp/pages/conditionlist/manageCondition.gsp
```diff
@@ -70,17 +71,18 @@ ${ui.includeFragment("coreapps", "patientHeader", [patient: patient])}
...
             <div id="status" class="horizontal">
                 <p>
-                    <input type="radio" id="status-1" class="condition-status" value="${ui.message('coreapps.conditionui.active.label')}" name="status" ng-model="condition.status"  ng-change="showEndDate()"/>
+                    <input type="radio" id="status-1" class="condition-status" value="ACTIVE" name="status" ng-model="condition.status"  ng-change="showEndDate()"/>
                     <label for="status-1">${ui.message('coreapps.conditionui.active.label')}</label>
                 </p>
                 <p>
-                    <input type="radio" id="status-2" class="condition-status" value="${ui.message('coreapps.conditionui.inactive.label')}" name="status" ng-model="condition.status"  ng-change="showEndDate()"/>
+                    <input type="radio" id="status-2" class="condition-status" value="INACTIVE" name="status" ng-model="condition.status"  ng-change="showEndDate()"/>
                     <label for="status-2">${ui.message('coreapps.conditionui.inactive.label')}</label>
```
mseaton, line 86: In testing with languages other than English, I found things to fail when trying to set inactive/active status.  I found this was due to the above, where the value of the status field was varying based on locale, rather than representing the enum value.

omod/src/main/webapp/pages/conditionlist/manageConditions.gsp
```diff
@@ -40,7 +41,7 @@ ${ui.includeFragment("coreapps", "patientHeader", [patient: patient])}
 <h2>${ui.message("coreapps.conditionui.manageConditions")}</h2>
 
 <div id="condition" ng-app="conditionsApp" ng-controller="ConditionsController"
-     ng-init="conditions = getConditions('${patient.uuid}')">
+     ng-init="conditions = getConditions('${patient.uuid}'); config = { dateFormat: '${ ui.getJSDateFormat() }', locale: '${ ui.getLocale().getLanguage() }'};">
     <div id="tabs">
```
mseaton, line 45: This is setting things up on the page so that the angular components have what they need from the UI framework (moment.js with localization and the configured date format and locale from the server) to support localized date formats.

omod/src/main/webapp/resources/scripts/conditionlist/controllers/condition.controller.js
```diff
@@ -8,7 +8,7 @@
         var onsetDatePicker;
         var startDateElement;
         var endDatePicker;
-        var endDateElement
+        var endDateElement;
```
mseaton, line 11: My build started failing and I tracked it down to lots of missing semi-colons.  It also didn't like the use of `let`, `const`, or other ES6+ syntaxes I tried to use.

omod/src/main/webapp/resources/scripts/conditionlist/controllers/condition.controller.js
```diff
@@ -119,19 +120,19 @@
...
-        }
+        };
 
         self.showEndDate = self.showEndDate || function () {
             var groups = document.getElementsByClassName("group");
             if ($scope.condition.status === INACTIVE_STATUS) {
                 groups[2].style.visibility = "visible";
                 if ($scope.condition.endDate) {
-                    groups[2].getElementsByTagName("input")[0].value = CommonFunctions.formatDate($scope.condition.endDate);
+                    endDatePicker.datetimepicker('setUTCDate', new Date($scope.condition.endDate));
```
mseaton, line 130: Here, I'm just invoking the same thing that is invoked on load.

omod/src/main/webapp/resources/scripts/conditionlist/controllers/condition.controller.js
```diff
@@ -153,27 +154,35 @@
...
+        };
+
+        self.toUTCDate = function(dateString) {
+            var utcDate = null;
+            if (dateString) {
+                utcDate = new Date(dateString);
+                utcDate.setTime(utcDate.getTime() + utcDate.getTimezoneOffset() * 60 * 1000);
+            }
+            return utcDate;
+        };
```
mseaton, line 175: The above is what fixes the issue in RA-1678 with an inconsistent date getting stored in the database from what was entered by the user.  Here, I adapted a solution I found that was undertaken to fix a similar issue in the program enrollment components of coreapps.  Specifically, I was inspired by this commit:  https://github.com/openmrs/openmrs-module-coreapps/commit/50b1c92e929fbcdeed93e0748d0ca7e0521b7826

omod/src/main/webapp/resources/scripts/conditionlist/controllers/conditions.controller.js
```diff
@@ -94,16 +95,24 @@
...
-        }
+        };
+
+        self.formatDate = self.formatDate || function(date) {
+            if (typeof date !== 'undefined' && date !== null) {
+                var locale = $scope.config.locale ? $scope.config.locale : 'en';
+                var format = $scope.config.format ? $scope.config.format : 'DD MMM YYYY';
+                return moment(date).locale(locale).format(format);
+            }
+        };
```
mseaton, line 106: This improves upon the previous formatDate function in that it uses the configured javascript date format to be consistent with other pages in the system and is localized.

ibacher (approved):

omod/src/main/webapp/resources/scripts/conditionlist/controllers/condition.controller.js
ibacher: Yeah, we build this with a very old version of NodeJS. I don't think `let` / `const` was supported until Node 8 and other bits (`import`) were much, much later.

omod/src/main/webapp/pages/conditionlist/manageCondition.gsp
mogoodrich: :)
Thanks!

omod/src/main/webapp/resources/scripts/conditionlist/controllers/condition.controller.js
mogoodrich: Not sure about this, @mseaton , will comment on the ticket.

PR #470, opened 2023-03-22
RA-2019: Update message properties for conditions

mseaton:
I'm pretty sure I'll need to force push these updates to TX before I commit them, right?  I'm getting an error currently using the tx client.  Maybe I have an old version.  I'd be curious if @ibacher  or @mogoodrich  or @dkayiwa are able to do a "tx pull" from the coreapps project.


ibacher:
When I run `tx pull` locally, I get this:

```
# Getting info about resources

OpenMRS.coreapps-module - Resource OpenMRS.coreapps-module does not exist
[##############################] (1 / 1)
```

Which... isn't terribly helpful.

mogoodrich:
Wow... it's been a long time, looks like I don't even have the tx client installed on this machine... :)

These are all changes you'd made manually?  Then, yes, you will have to force push them up, but you should check first to make sure that there aren't any changes made in Transifex that nobody has pulled in, otherwise you'll blow them away with your force push.

dkayiwa:
Are the merge conflicts caused by the transfix pull?

mseaton:
OK, I'm closing out this PR, as I ended up doing things through transifex and working through upgrading it to work again.  I'll comment on the ticket.

PR #471, opened 2023-03-24
RA-1678 - Update on timezone fix that moves date interpretation to th…

mseaton:
…e server endpoint

This PR corresponds with https://github.com/openmrs/openmrs-module-emrapi/pull/211, and reverts back to the original/simpler functionality that submits the entered date as-is to the server, with no manipulation.


mogoodrich (commented):
LGTM!

PR #472, opened 2023-03-29
RA-1678: Fix display of dates to ignore time and timezone components.

mseaton:


omod/src/main/webapp/resources/scripts/conditionlist/controllers/conditions.controller.js
```diff
@@ -102,7 +102,7 @@
             if (typeof date !== 'undefined' && date !== null) {
                 var locale = $scope.config.locale ? $scope.config.locale : 'en';
                 var format = $scope.config.format ? $scope.config.format : 'DD MMM YYYY';
-                return moment(date).locale(locale).format(format);
+                return moment(date.substring(0, 10)).locale(locale).format(format);
```
mogoodrich, line 105: Looks correct, for reference here is the code I have been using, with Brandon's comments on why it is horrible. :)

https://github.com/openmrs/openmrs-module-uicommons/blob/master/omod/src/main/webapp/resources/scripts/filters/serverDate.js

It might be worth looking into this function that Brandon references, though I think I give it a quick look and it didn't work for our case?  (But I 100% could be wrong about that)

mseaton: Yes, I did look at those @mogoodrich , and this ended up being the cleanest solution, and also one that doesn't result in tons of annoying console log messages that tell me now to use it :)

mogoodrich: I thought I finally disabled those messages, fwiw...

PR #475, opened 2023-12-05
Program display should not overflow into date display.

mseaton:
Truncates the program name with ellipses (...) if the program name exceeds 20 characters (adding the full program name as a tooltip to view on hover), to ensure that it does not overflow into the dates.

Example:
image


cioan (approved):
Looks great, @mseaton ! Thanks!

mogoodrich (approved):
Sweet!  I was struggling with CSS without luck, but is arguably a better solution anyway...

dmdesimone:
agree, nice @mseaton

PR #477, opened 2023-12-20
UHM-7623, display error when failing to save patient

cioan:
@mogoodrich , is there anything else we need to do display this validation error? I just printed the error message:

Screenshot 2023-12-20 at 4 51 50 PM


mogoodrich (commented):
Looks great, thanks @cioan! I would just add an overall message like "Unable to mark patient as deceased: " + then display the underlying error.

cioan:
Thanks @mogoodrich ! I have updated the error message.

PR #481, opened 2024-03-18
O3-2897: Add support for use of the concept name on  the global property `concept.causeOfDeath`

icrc-loliveira:
Issue: https://openmrs.atlassian.net/browse/O3-2897


icrc-loliveira:
Hello @ibacher,
Can you please review this PR.
I have used the function `getConcept` since the function `getConceptByReference()` is not available in the `openmrs-core version` in use  in the `coreapps`

ibacher (approved):

PR #482, opened 2024-06-06
RA-2039 - Update minimum OpenMRS version to 2.2.1 and major version change to 2.x

mseaton:
This PR is a result of having made backwards-incompatible changes to EMR API 2.x (specifically changing the return type of some methods in AccountDomainWrapper to make providermanagement an optional dependency), and the need to support a version of Core Apps that is compatible with this new API.  Main changes are as follows:

* Change module version to 2.0.0
* Upgraded minimum version of OpenMRS to 2.2.1, and increased versions of several dependencies to later versions
* Cleaned up and fixed issues in the poms
* Removed api-pre2.2 submodule (was only previously loaded if version was < 2.2)
* Removed the compatibility classes for AddressSupport, NameSupport, and EncounterService in favor of just using the native methods present in 2.2+
* Removed a few public methods that were not used (VisitTypeHelper)
* Moved several classes (all of htmlformentry + VisitContextModel + the activator) into the omod project, as they have web-specific dependencies
* Added an aware-of dependency on legacyui and added some adapter code for it to QuickVisitFragmentController, which uses functionality that was moved out of core and into legacyui years ago
* Renamed some of the htmlformentry tests to reflect what they represent now that the code is consolidated
* Fixed the EditProviderPageController to reflect changes in emrapi 2.0.0 (the main need for this ticket)

Still need to do some additional testing, but this is ready for an overall code review.


api/src/test/java/org/openmrs/module/coreapps/htmlformentry/EncounterDiagnosesElementTest.java
mogoodrich, line 1: Same question here as with the other encounter diagnosis stuff

api/src/main/java/org/openmrs/module/coreapps/utils/VisitTypeHelper.java
```diff
@@ -189,76 +181,4 @@ public Map<Integer, Map<String, Object>> getVisitColorAndShortName(List<VisitDom
 		}
 		return visitsWithAttr;
 	}
-
-	/**
-	 * Sets encounters based on visit type and login location
```
mogoodrich, line 194: These are the ones that aren't used?   Looks like they were added for this: https://openmrs.atlassian.net/browse/RA-977, but that seems to all be tied to stuff in this module (I was trying to check to see if it was used by a dependent module).
tl;dr... yeah, if these aren't used within coreapps, good to remove

api-2.2/src/test/java/org/openmrs/module/htmlformentry/EncounterDiagnosesTagHandlerComponentTest.java
mogoodrich, line 1: I assume all the encounter diagnosis stuff was moved, not deleted, but git is picking it up as a delete-and-recreate?

mogoodrich (approved):
Generally looks good to me.  It was a little difficult to review the EncounterDiagnosis stuff, but I assume that was just moved and  tweaked?

Everything in your explanation makes sense.

api-2.2/src/test/java/org/openmrs/module/htmlformentry/EncounterDiagnosesTagHandlerComponentTest.java
mseaton: Yes, this was moved.  The EncounterDiagnosis stuff was a little confusing, but this test was not deleted, just moved and renamed.

api/src/main/java/org/openmrs/module/coreapps/utils/VisitTypeHelper.java
mseaton: Yes.  Basically, these methods are trying to create an encounter, and to set a provider on an encounter, and this is no longer possible in 2.x because we need to know what encounter role to use when assigning the provider.  But then I realized that these methods were not actually used, so I could remove them and not have to figure out how to overcome this incompatibility, so I did.  If something downstream was using them, we can adapt there I hope.

api/src/test/java/org/openmrs/module/coreapps/htmlformentry/EncounterDiagnosesElementTest.java
mseaton: Yes, same answer.  The number 1 thing I try to do with a major refactor like this is to preserve all existing unit tests, as that is the best way to make sure I didn't break anything.

omod/src/test/java/org/openmrs/module/coreapps/htmlformentry/EncounterDispositionTagHandlerComponentTest.java
```diff
@@ -192,16 +190,15 @@ public Map<String, Object> getFormEntrySessionAttributes() {
 
             @Override
             public String[] widgetLabels() {
-                return new String[] { "Date:", "Location:", "Provider:", "Disposition:", "Encounter Type:" };
```
dkayiwa, line 195: Could you remind me why we had to drop Encounter Type?

mseaton: I thought at one point it was leading to a test failure.  I think it was a red herring, but the tests pass without it, and it isn't what is being tested, so I just left it out.

dkayiwa (approved):

PR #483, opened 2024-06-06
RA-2040 - Remove dependence on rubygems for sass

mseaton:
See:  https://openmrs.atlassian.net/browse/RA-2040

This takes the same approach as we have used in other modules.  I borrowed heavily from the setup in uicommons and pihcore.


mogoodrich (approved):
I don''t remember this 100% @mseaton , I but I assume this is just copying from another module were we (Brandon, IIRC?) removed the dependency on rubygems.  Sounds great to me.

PR #484, opened 2024-06-06
Core Apps - Update karma tests to build with ChromeHeadless

mseaton:
See:  https://openmrs.atlassian.net/browse/RA-2041

This is based off of a series of PRs / branches, so would currently need https://openmrs.atlassian.net/browse/RA-2039 and https://openmrs.atlassian.net/browse/RA-2040 to be committed prior to this.

It's just easier to review and break up the work this way.


mogoodrich (approved):
I forget the whole history re: this, defer to you @mseaton .  thanks!

dkayiwa (approved):

PR #489, opened 2024-09-18
SL-671: display the active program enrollmenets

cioan:
@mogoodrich , I was able to include a table with the current program enrollment in this new page:

Screenshot 2024-09-18 at 2 03 47 PM

It was nice that the statistics numbers matched the number of the rows in this new table. :-)

### DIFF
--- a/omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
@@ -1,0 +1,82 @@
+package org.openmrs.module.coreapps.page.controller.summarydashboard;
+
+import org.apache.commons.lang.StringUtils;
+import org.codehaus.jackson.JsonNode;
+import org.codehaus.jackson.node.ObjectNode;
+import org.openmrs.PatientProgram;
+import org.openmrs.Program;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.appframework.context.AppContextModel;
+import org.openmrs.module.appframework.domain.AppDescriptor;
+import org.openmrs.module.appframework.domain.Extension;
+import org.openmrs.module.appframework.service.AppFrameworkService;
+import org.openmrs.module.appui.UiSessionContext;
+import org.openmrs.module.coreapps.CoreAppsConstants;
+import org.openmrs.module.coreapps.CoreAppsProperties;
+import org.openmrs.module.emrapi.EmrApiProperties;
+import org.openmrs.ui.framework.annotation.SpringBean;
+import org.openmrs.ui.framework.page.PageModel;
+import org.openmrs.ui.framework.page.Redirect;
+import org.springframework.web.bind.annotation.RequestParam;
+
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.Date;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+public class ProgramEnrollmentsPageController {
+
+    public Object controller(@RequestParam("app") AppDescriptor app,
+                             @SpringBean EmrApiProperties emrApiProperties,
+                             @SpringBean("coreAppsProperties") CoreAppsProperties coreAppsProperties,
+                             @SpringBean("appFrameworkService") AppFrameworkService appFrameworkService,
+                             PageModel model,
+                             UiSessionContext sessionContext) {
+
+        if (!Context.hasPrivilege(CoreAppsConstants.PRIVILEGE_SUMMARY_DASHBOARD)) {
+            return new Redirect("coreapps", "noAccess", "");
+        }
+
+        model.addAttribute("app", app);
+
+        AppContextModel contextModel = sessionContext.generateAppContextModel();
+        model.addAttribute("appContextModel", contextModel);
+
+        List<Extension> includeFragments = appFrameworkService.getExtensionsForCurrentUser(app.getId() + ".includeFragments", contextModel);
+        Collections.sort(includeFragments);
+        model.addAttribute("includeFragments", includeFragments);
+
+        List<Extension> firstColumnFragments = appFrameworkService.getExtensionsForCurrentUser(app.getId() + ".firstColumnFragments", contextModel);
+        Collections.sort(firstColumnFragments);
+        model.addAttribute("firstColumnFragments", firstColumnFragments);
+
+        String programUuid = null;
+        ObjectNode config = app.getConfig();
+        if (config != null) {
+            for (Iterator<Map.Entry<String, JsonNode>> fields = config.getFields(); fields.hasNext();) {
+                Map.Entry<String, JsonNode> field = fields.next();
+                if (StringUtils.equalsIgnoreCase(field.getKey(), "program") ){
+                    programUuid = field.getValue().getTextValue();
+                }
+            }
+        }
+        if (programUuid != null ) {
+            Program program = Context.getProgramWorkflowService().getProgramByUuid(programUuid);
+            if (program != null ) {
+                Date today = Calendar.getInstance().getTime();
+                // retrieve all program enrollments still active as now
+                List<PatientProgram> patientPrograms = Context.getProgramWorkflowService().getPatientPrograms(null, program, null, today, today, null, false);
+                for (PatientProgram patientProgram : patientPrograms) {
+                    patientProgram.getDateEnrolled();
+                    patientProgram.getDateCompleted();
+                }
+                model.addAttribute("programEnrollments", patientPrograms);
+            }
+        }
+        model.addAttribute("dashboardUrl", coreAppsProperties.getDashboardUrl());
+        model.addAttribute("primaryIdentifierType", emrApiProperties.getPrimaryIdentifierType());
+        return null;
+    }
+}
--- a/omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
@@ -68,10 +68,6 @@ public class ProgramEnrollmentsPageController {
                 Date today = Calendar.getInstance().getTime();
                 // retrieve all program enrollments still active as now
                 List<PatientProgram> patientPrograms = Context.getProgramWorkflowService().getPatientPrograms(null, program, null, today, today, null, false);
-                for (PatientProgram patientProgram : patientPrograms) {
-                    patientProgram.getDateEnrolled();
-                    patientProgram.getDateCompleted();
-                }
                 model.addAttribute("programEnrollments", patientPrograms);
             }
         }
--- a/omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
+++ b/omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
@@ -64,32 +64,35 @@ public class ProgramEnrollmentsPageController {
                 }
             }
         }
+        Set<ProgramWorkflow> workflows = null;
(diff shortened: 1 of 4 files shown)


omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
```diff
@@ -0,0 +1,78 @@
...
+                    programUuid = field.getValue().getTextValue();
+                }
+            }
+        }
+        if (programUuid != null ) {
+            Program program = Context.getProgramWorkflowService().getProgramByUuid(programUuid);
+            if (program != null ) {
+                Date today = Calendar.getInstance().getTime();
+                // retrieve all program enrollments still active as now
+                List<PatientProgram> patientPrograms = Context.getProgramWorkflowService().getPatientPrograms(null, program, null, today, today, null, false);
```
mseaton, line 70: I worry about how this will perform for all programs.  I don't really have a better solution in the short term unless we can achieve server-side paging restfully, but just something to watch out for as some programs can have many hundreds or thousands of active patients. I also don't know if we would want to limit this in some way around enrollment location - in our Malawi case, you likely want to see enrollments based on location.

omod/src/main/webapp/pages/summarydashboard/programEnrollments.gsp
```diff
@@ -0,0 +1,91 @@
...
+		<tbody>
+			<% programEnrollments.each { e ->
+			%>
+			<tr>
+			<td class="name-link"><a href="${ ui.urlBind("/" + contextPath + dashboardUrl, [ patientId: e.patient.patientId ]) }">${ ui.format(e.patient.personName)}</a></td>
+				<td>${ e.patient.getPatientIdentifier(primaryIdentifierType.id) }</td>
+				<td>${ ui.format(e.dateEnrolled)}</td>
+				<td>
+					<% e.currentStates.eachWithIndex { state, index -> %>
+						${ ui.format(state.state.concept) }${ e.currentStates.size() - index > 1 ? ", " : ""}
```
mseaton, line 72: Is this what is in the requirements?  I wouldn't expect us to be showing all of the states here like this.  I could see potentially having one column per workflow in the program, and then showing the current state (if any) for that workflow for each patient.  That would make more sense to me.

omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
```diff
@@ -0,0 +1,78 @@
...
+        AppContextModel contextModel = sessionContext.generateAppContextModel();
+        model.addAttribute("appContextModel", contextModel);
+
+        List<Extension> includeFragments = appFrameworkService.getExtensionsForCurrentUser(app.getId() + ".includeFragments", contextModel);
+        Collections.sort(includeFragments);
+        model.addAttribute("includeFragments", includeFragments);
+
+        List<Extension> firstColumnFragments = appFrameworkService.getExtensionsForCurrentUser(app.getId() + ".firstColumnFragments", contextModel);
+        Collections.sort(firstColumnFragments);
+        model.addAttribute("firstColumnFragments", firstColumnFragments);
```
mogoodrich, line 53: Are we only having a first column on this dashboard, not a second? 

I don't think we necessarily need to support two columns, but,in that case, it  might make sense to rename this something else rather than "firstColumn", though admittedly a great name doesn't spring to mind.

omod/src/main/webapp/pages/summarydashboard/programEnrollments.gsp
```diff
@@ -0,0 +1,91 @@
...
+
+<div class="clear"></div>
+<div class="container">
+	<div class="dashboard clear">
+		<!-- only show the title div if a title has been defined in the messages.properties -->
+		<% if (ui.message(app.id + ".custom.title") != app.id + ".custom.title") { %>
+		<div class="title">
+			<h3>${ ui.message(app.id + ".custom.title") }</h3>
+		</div>
+		<% } %>
```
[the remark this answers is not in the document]
mogoodrich: Is  what @mseaton  said true?  I agree we wouldn't want to take first column fragments designed for another page/app and display them here, but it looks like the columns are defined by the name of this particular app, though I could be forgetting how it all works or missing something obvious.  fyi @cioan

omod/src/main/webapp/pages/summarydashboard/programEnrollments.gsp
```diff
@@ -0,0 +1,91 @@
...
+	<h3>${ ui.message("coreapps.currently.enrolled") }:</h3>
+</div>
+<div>
+	<table id="enrollmentsTable" style="width: 100%">
+		<thead>
+			<tr>
+				<th>${ ui.message("Patient.name") }</th>
+				<th>${ ui.message("coreapps.emr.id") }</th>
+				<th>${ ui.message("coreapps.enrollment.startDate") }</th>
+				<th>${ ui.message("coreapps.treatment.state") }</th>
```
mogoodrich, line 60: This likely shouldn't be hardcoded to "Treatment State", as that is a specific workflow.  If we are displaying *all* current states in a single column, we'd need to name this something like "State(s)", but I agree with @mseaton 's comment below that it would likely be better to be column-per-workflow.

omod/src/main/webapp/pages/summarydashboard/programEnrollments.gsp
mogoodrich: Yeah, the mockup is based around a program with a single workflow, Treatment State, and doesn't address how to handle programs with multiple workflows.  I agree with @mseaton  that we should have a column per workflow, displaying the current state for that workflow in the column.  Happy to discuss more tomorrow.

omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        model.addAttribute("app", app);
+
+        AppContextModel contextModel = sessionContext.generateAppContextModel();
+        model.addAttribute("appContextModel", contextModel);
+
+        List<Extension> includeFragments = appFrameworkService.getExtensionsForCurrentUser(app.getId() + ".includeFragments", contextModel);
+        Collections.sort(includeFragments);
+        model.addAttribute("includeFragments", includeFragments);
```
mogoodrich, line 49: It looks like we aren't using these "includeFragments" in the page itself, we are only using the "firstColumnFragments"? Or am I missing them.  If we aren't using them, let's remove this block?

mogoodrich (changes requested):
A few comments/requests for changes... primarily, I think it makes sense to do the column-per-workflow that @mseaton  described, but, otherwise, looking good, thanks @cioan !

omod/src/main/webapp/pages/summarydashboard/programEnrollments.gsp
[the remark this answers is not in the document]
cioan: Thanks @mogoodrich ! I am not sure how all this works either. I thought I could just use the current CustomAppLoaderFactory configuration to display the already configured program statistics widget here. But since this is a new page, I could just remove this type of on the fly configuration, and just hard code the program statistics widget at the top of the page and the table of enrollments underneath. Would that be better? That would mean removing all the existing program summary configuration from the CustomAppLoaderFactory because it would be pointless if we hard coded in the new page?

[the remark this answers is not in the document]
mogoodrich: @cioan  do you have a branch with the CALF with the changes you made?  It will likely be easier for me to review/understand from that.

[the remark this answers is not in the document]
cioan: @mogoodrich : https://github.com/PIH/openmrs-module-pihcore/pull/524

omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
mogoodrich: So, yes, let's remove this block (47-49) since we aren't using "includeFragments" in the rendered model.

omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
mogoodrich: This is fine, let's just keep this a "firstColumn" for now.

mogoodrich (changes requested):
Thanks @cioan , so generally this looks good, just:
* removing that block of code in the controller that I don't believe is doing anything (see my comment)
* (a little more work) Making a column-per-workflow-state

cioan:
Thanks @mogoodrich ! I have updated the PR to address the observations you made. Thanks!

omod/src/main/java/org/openmrs/module/coreapps/page/controller/summarydashboard/ProgramEnrollmentsPageController.java
```diff
@@ -0,0 +1,97 @@
...
+                    for(ProgramWorkflow workflow : workflows) {
+                        PatientState currentState = patientProgram.getCurrentState(workflow);
+                        simpleObjects.add(SimpleObject.create(
+                                "patientUuid", patient.getUuid(),
+                                "patientId", patient.getPatientId(),
+                                "personName", patient.getPersonName().getFamilyName() + ", " + patient.getPersonName().getGivenName(),
+                                "emrId", patient.getPatientIdentifier(emrApiProperties.getPrimaryIdentifierType()),
+                                "dateEnrolled", patientProgram.getDateEnrolled(),
+                                "dateCompleted", patientProgram.getDateCompleted(),
+                                "treatmentState", (currentState != null) ? currentState.getState().getConcept().getName(Context.getLocale()) : ""
```
mogoodrich, line 86: Is there a screen shot for this, and did we test it out on a program with multiple workflows?  Maybe I'm mis-reading, but won't this end up with a row-per-patient-program-workflow, not a row per patient-program? I was thinking that multiple workflows would result in multiple columns, not rows, happy to discuss at standup.

cioan: @mogoodrich , here is the screenshot, which reflects the mockup (https://pihemr.atlassian.net/browse/SL-671 )

Screenshot 2024-09-20 at 10 32 35 AM

I don't understand how multiple workflows would be displayed in multiple columns? Do you have an example of a program with multiple workflows? Just curious how would that look in the legacy Admin UI, or any other UI if we already have something like this anywhere?

mogoodrich: Looks like the HIV and Oncology programs both have 2 states, though it looks like they aren't in KGH.  Here's HIV:

image

So in this case, there are two workflows, "HIV Treatment Status" and "PMTCT Status".  So, ideally, there would be a column for each in the display.

I *think* the current code would create two rows for this program, one with a treatment state of "On ART" and one of "Patient Pregnant", which is definitely not what we want.

We could get away with just combined the two states into a single column, but we've want to name it something generic like "State" instead of "Treatment State".

cioan: Thanks @mogoodrich ! I got it working with each workflow current state in a separate column. And the translation of the column name was easy by just adding this to the messages.properties:
+ui.i18n.ProgramWorkflow.name.9a3f8252-1588-4f7b-b02c-9e99c437d4ef=Treatment State

mogoodrich (approved):
LGTM, thanks @cioan !

cioan:
Thanks @mogoodrich !

PR #492, opened 2024-09-30
Add a loader when searching for a patient with fingerprints

louidorjp:
Please remove this PR. 
FYI: @mogoodrich , @mseaton

### DIFF
--- a/omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
+++ b/omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
@@ -50,7 +50,7 @@ function PatientSearchWidget(configuration){
     });
     tableHtml += '</tr></thead><tbody></tbody></table>';
 
-    var spinnerImage = '<span><img class="search-spinner" src="'+emr.resourceLink('uicommons', 'images/spinner.gif')+'" /></span>';
+    var spinnerImage = '<span id="spinner-image"><img class="search-spinner" src="'+emr.resourceLink('uicommons', 'images/spinner.gif')+'" /></span>';
 
     jq('#'+config.searchResultsDivId).addClass("table-responsive").append(tableHtml);
     var input = jq('#'+config.searchInputId);
@@ -210,6 +210,7 @@ function PatientSearchWidget(configuration){
                 .then(function (data) {
                     if (data && data.results && data.results.length > 0) {
                         updateSearchResults(data.results);
+                        jq('#spinner-image').remove();
                     }
                 })
             deferredList.push(deferred);
@@ -220,13 +221,16 @@ function PatientSearchWidget(configuration){
             .done(function () {
                 if (searchResultsData.length == 0) {
                     displayNoMatchesFound();
+                    jq('#spinner-image').remove();
                 }
                 else if (searchResultsData.length == 1) {
                     selectRow(0);  // auto-select if one match, may want to make this configurable
+                    jq('#spinner-image').remove();
                 }
             })
             .fail(function (jqXHR) {
                 failSearch();   // TODO is this what we want here?
+                jq('#spinner-image').remove();
             });
 
     }
@@ -791,6 +795,7 @@ function PatientSearchWidget(configuration){
     jq('#patient-search-form').on('search:no-matches', function(event, identifier) {
         clearSearch();
         displayNoMatchesFound();
+        jq('#spinner-image').remove();
     })
 
     jq('#patient-search-form').on('search:placeholder', function(event, placeholder) {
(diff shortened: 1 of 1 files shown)


omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
```diff
@@ -804,4 +809,27 @@ function PatientSearchWidget(configuration){
...
+
+
+    // Store the original XMLHttpRequest open function to preserve its functionality
+    const originalXHR = window.XMLHttpRequest.prototype.open;
+
+    // Override the XMLHttpRequest open function to intercept specific requests
+    window.XMLHttpRequest.prototype.open = function (method, url) {
+
+        // Check if the requested URL contains '/openmrs/registrationapp/biometrics/biometrics/search.action'
+        if (url.includes('/openmrs/registrationapp/biometrics/biometrics/search.action')) {
```
mseaton, line 821: This should not check for the leading "/openmrs", as the application may be deployed with a different webapp name than openmrs.

omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
```diff
@@ -50,7 +50,7 @@ function PatientSearchWidget(configuration){
     });
     tableHtml += '</tr></thead><tbody></tbody></table>';
 
-    var spinnerImage = '<span><img class="search-spinner" src="'+emr.resourceLink('uicommons', 'images/spinner.gif')+'" /></span>';
+    var spinnerImage = '<span id="spinner-image"><img class="search-spinner" src="'+emr.resourceLink('uicommons', 'images/spinner.gif')+'" /></span>';
```
mseaton, line 53: Do we need to add this id?  Can't we just use a class-based selector with "search-spinner" ?

omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
```diff
@@ -804,4 +809,27 @@ function PatientSearchWidget(configuration){
...
+
+    // Override the XMLHttpRequest open function to intercept specific requests
+    window.XMLHttpRequest.prototype.open = function (method, url) {
+
+        // Check if the requested URL contains '/openmrs/registrationapp/biometrics/biometrics/search.action'
+        if (url.includes('/openmrs/registrationapp/biometrics/biometrics/search.action')) {
+
+            jq('#patient-search-form').after(spinnerImage);
+
+            // Apply CSS styles to position the spinner
```
mseaton, line 825: This seems a little too specific to one particular form where this widget is used.  This widget is potentially used in a lot of different places, and it may not have the same positioning on the page, etc.  But I need to refresh myself on how all of this is wired together.

mogoodrich (commented):
Thanks @louidorjp ! See my comments, as I said, it's been a while since I 've touched this code and I don't have a way to run the fingerprint stuff right now on my laptop, so hopefully my comments make sense.  Happy to discuss further, take care!

louidorjp:
@mseaton, @mogoodrich  In the fingerprintSearchController, within the function that initiates the fingerprint scan, when the scan is successful, the search:clear event is triggered via the command
`jq('#patient-search-form').trigger('search:clear');`
 As a result, a loader is displayed when the event handler listens for this custom **search:clear** event.

Ref: https://github.com/openmrs/openmrs-module-registrationapp/blob/master/omod/src/main/webapp/resources/scripts/biometrics/fingerprintSearchController.js#L84

omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
louidorjp: instead of displaying it every time a “searchOnIdentifier” is triggered, I display it when the search:clear event is triggered

omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
```diff
@@ -770,6 +774,7 @@ function PatientSearchWidget(configuration){
     // see https://issues.openmrs.org/browse/RA-1404 for potential use cases
     jq('#patient-search-form').on('search:clear', function() {
         clearSearch();
+        jq('#patient-search-form').after(spinnerImage);
```
mogoodrich, line 777: What do you think about instead adding this line here, we make it more explicit the and add a new "start" event, like:

```
jq('patient-search-form').on('search:start', function() {
            jq('#patient-search-form').after(spinnerImage);
}
```

and then in the fingerprintSearchController we do:

```
jq('#patient-search-form').trigger('search:clear')

jq('#patient-search-form').trigger('search:start')
```

This makes it a little explicity that "clearing" the search box and rendering the spinner are two separate events that we just happen to want to do at the same time.

omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
louidorjp: I agree with you @mseaton . I've removed the css properties.

mogoodrich (approved):
Great, much shorter, looks good to me! I did add one suggestion, let me know what you think....

omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
louidorjp: @mogoodrich yes, I had this idea, I wanted to propose this to you.

mogoodrich (approved):

PR #493, opened 2024-10-03
SL-734: sequentially chain the REST calls

cioan:
@mogoodrich , I finally figured out the problem with this widget, and why sometimes it did not show any data. The problem was that all the REST calls were executed asynchronously with no way of controlling the order of those REST calls. Therefore, sometimes the obs REST call was returning before the programenrollment, and the widget was configured to display only obs during a certain program enrollment but we had no enrollment date because the REST call did not return before the obs rest.
So, now I chained those REST calls to execute sequentially, the grapah loads correctly every time:

Screenshot 2024-10-03 at 10 07 39 AM


mogoodrich (approved):
LGTM, thanks @cioan !

PR #495, opened 2024-10-04
RA-2053: Core Apps: Encounter Diagnoses widget should not allow entry…

mogoodrich:
… of non-coded diagnosis that is all whitespace


mseaton (approved):
Defer to you on the correctness of the regex.  LGTM

PR #496, opened 2024-11-04
RA-2062 - Patient Search should support a configurable search handler

mseaton:


cioan (approved):
LGTM, thanks @mseaton !

PR #497, opened 2024-11-07
RA-2062 - Patient Search search handler should not be set by default

mseaton:


omod/src/main/resources/config.xml
```diff
@@ -225,7 +225,7 @@
 
     <globalProperty>
         <property>coreapps.patientSearchHandler</property>
-        <defaultValue>patientByIdentifier</defaultValue>
+        <defaultValue></defaultValue>
         <description>
             Indicates which patient search handler implementation should be used by the patient search widget
```
mogoodrich, line 230: Update this to say something like "Set this to override the built-in search handlers with a custom search handler", or something like that?

mogoodrich (approved):
LGTM, though would be good to update the description of the GP, see my comments.

omod/src/main/resources/config.xml
mseaton: Done

PR #502, opened 2025-01-28
RA-2069 - Allow configuring death reasons from concept sets

mseaton:
See:  https://openmrs.atlassian.net/browse/RA-2069

[Screencast from 2025-01-28 17-30-28.webm](https://github.com/user-attachments/assets/167f13e0-8ffd-417e-a73a-28fa37257a63)


mogoodrich (approved):
LGTM, thanks @mseaton !

PR #508, opened 2025-05-02
UHM-8591: remove any potential null values from the search results

cioan:
@mogoodrich , here is just a small fix that removes null objects from the search results that the patient REST could return.

### DIFF
--- a/omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
+++ b/omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
@@ -315,7 +315,7 @@ function PatientSearchWidget(configuration){
     var updateSearchResults = function(results){
         var dataRows = [];
         if(results){
-            //first remove any null objects from the results
+            //first remove any null objects and then remove the duplicates from the results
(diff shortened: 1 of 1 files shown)


omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
```diff
@@ -315,7 +315,8 @@ function PatientSearchWidget(configuration){
     var updateSearchResults = function(results){
         var dataRows = [];
         if(results){
-            var results = removeDuplicates(results);
+            //first remove any null objects from the results
```
mogoodrich, line 318: nit, at the end of the day, the line below removes null objects and then removes duplicates, correct?  (the way the comment is worded I read it to mean all the line does is remove null elements...
0

mogoodrich (approved):
LGTM, thanks @cioan , nit comment on the comment... :)

omod/src/main/webapp/resources/scripts/patientsearch/patientSearchWidget.js
cioan: Thanks @mogoodrich ! I have updated the comments.

PR #511, opened 2025-05-23
RA-2079: add patientPrograms to the AppContextModel

cioan:


mseaton (commented):
This looks good to me @cioan .  The only thing is that this also introduces the breaking change of renaming `patientPrograms` to `programs` and then changing the meaning of `patientPrograms`.  So we should indicate that in the commit message and ticket (and potentially bump up the version in the pom to the next major release version?)

PR #533, opened 2026-05-11
RA-2107: AppContextModel: add new "hasAnyActiveVisit" to model

mogoodrich:


mogoodrich:
Was trying to come up with a good name for this property... use case is we only want the "Start Visit" button to appear if the patient does not have an active visit at *any* location, so we need a way to figure this out.

Took the opportunity to have Claude add tests for all of this as well.

cioan (approved):
Looks good! Thanks @mogoodrich !

mseaton (approved):
Looks fine to me.  That is a disproportionally massive number of tests for this feature!  :)  I assume you wrote all of those yourself.

mogoodrich:
Thanks @cioan  @mseaton ... yeah, there were *no* tests around this class at all, so I took the opportunity to have Claude create them.  Really helpful.

PR #534, opened 2026-05-18
RA-2108: CoreApps: Have way to prevent adding a past visit at a diffe…

mogoodrich:
…rent visit location


cioan (approved):
Looks good! Thanks @mogoodrich !

mseaton (approved):

PR #535, opened 2026-05-20
update various JS dependencies/libraries

mogoodrich:


.github/workflows/build.yml
```diff
@@ -19,7 +19,6 @@ jobs:
       MAVEN_REPO_API_KEY: ${{ secrets.MAVEN_REPO_API_KEY }}
 
   owasp-dependency-check:
-    if: ${{ github.event_name != 'pull_request' }}
```
mogoodrich, line 22: I just did this so the OWASP tests would run against the branch... we can remove, but is there any harm in doing this?

